PR #23, opened 2020-09-16
21 add DICOM format in the LoadImage tutorial

Nic-Ma:
This PR updated the `load medical image` tutorial to support DICOM format.


wyli (commented):
thanks @Nic-Ma, a few questions:
- could you add a multi-channel example from multiple DICOM files
- does it support loading a folder of dcm files
- could you change the example shape `(128 128 128)` to something like (128, 128, 96)? so that we could identify the shape issues easily if there's any

Nic-Ma:
Hi @wyli ,

Thanks for your review.
I updated the PR according to your review, `loading DICOM in a folder` is another PR in MONAI core repo.
Could you please help review it again?

Thanks.

Nic-Ma:
Hi @wyli and @ericspod ,

As we fixed ITK shape issue in PR: https://github.com/Project-MONAI/MONAI/pull/1045, updated the notebook here.
This PR is ready for review now.

Thanks.

wyli (approved):

Issue #33, opened 2020-09-28
double check all examples/tutorials


wyli:
since there're breaking changes since v0.2, we need to rerun and double-check all the examples/tutorials for v0.3

PR #37, opened 2020-10-01
update lightning/catalyst tutorials

wyli:
part of #33 
adds a `.gitignore`
- pytorch lightning tutorial: 
    - trainer `show_progress_bar=False,` is deprecated, should be removed
    - it uses much more RAM compared with a vanilla training, perhaps there's a compatibility issue with CacheDataset?
- unet_segmentation_3d_catalyst: update a URL

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,132 @@
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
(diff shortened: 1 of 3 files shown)


Nic-Ma (approved):
Looks good to me.
Thanks.

PR #47, opened 2020-10-20
add occlusion sensitivity to 2d and 3d classifications

rijobro:
Example of 2D occlusion sensitivity. Requires https://github.com/Project-MONAI/MONAI/pull/1099


rijobro:
Now that https://github.com/Project-MONAI/MONAI/pull/1099 has been merged, this can go in, too.

The only problem is that the occlusion code isn't part of a tagged release yet, so presumably won't be available if the notebook contains, e.g.,:
```
%pip install -q monai
```
which the 2D classification tutorial does...

wyli:
> Now that [Project-MONAI/MONAI#1099](https://github.com/Project-MONAI/MONAI/pull/1099) has been merged, this can go in, too.
> 
> The only problem is that the occlusion code isn't part of a tagged release yet, so presumably won't be available if the notebook contains, e.g.,:
> 
> ```
> %pip install -q monai
> ```
> 
> which the 2D classification tutorial does...

that's a good point, one option would be pip install the tested commit directly
```
%pip install git+git://github.com/project-monai/monai.git@8169061cab567ec660373baa17136b916b00ab04
```
but if the notebooks depend on different commits this repo becomes messy... @ericspod @Nic-Ma

rijobro:
I would be in favour of removing the `pip install monai` altogether, it gets in the way of my local copy (not pip installed)!

ericspod:
We do want to include tutorials into the future based on experimental features or at least those not on a current version. If we leave the `pip install monai` line as is we can put a note stating the version must at least be at a given commit or later, eventually when 0.4.0 is out we can modify the notebook to remove that. The alternative thing to do is like what would work for rijobro is not have the pip install anywhere and rely on the user to install things, I feel this is more common in the example notebooks I've seen across github and elsewhere rather than hard-coding the install process.

rijobro:
I've moved the discussion of pip installs to issue #50, and in the meantime removed them from the notebook in this PR. 

As such, think this is good to be merged.

wyli (approved):

PR #52, opened 2020-10-30
a baseline method with unet

wyli:


Nic-Ma (approved):
Looks good to me.
Thanks.

PR #111, opened 2021-01-21
Correct rotate_range and scale_range

Ordgod:
The parameters of rotate_range and scale_range should be 3 values which tune the transform extent on 3 dimensions. And it's better to use 3 positive values to avoid  ambiguous meanings.


wyli (approved):
thanks!

PR #120, opened 2021-02-01
Update the max line length of autopep8 to match what's in the main repo

ericspod:
This is a slight change to set the max line length to 120 like in the main repo.


rijobro:
Think you'd also need to add `--max-line-length 120` to line containing `flake8 - --show-source`.

Could you run `runner.sh --autofix --no-run` and see if any files get changed? Not sure if `autopep8` will remove line breaks for anything between 79 and 120 characters long, although I can't imagine that it would.

ericspod:
Updated. This seems to work fine.

Nic-Ma (approved):
Looks good to me, @rijobro could you please help double confirm as you are expert on this topic.

Thanks in advance.

PR #126, opened 2021-02-22
Update for DynUNet's changes and fix get_feature_maps issue

yiheng-wang-nv:
Fixes #106 and #128 .

**Description**
Due to the forward function's changes of DynUNet, this commit
update the corresponding places, as well as the loss calculation
for trainer.

In addition, the DiceCEloss has been implemented in MONAI, thus the
self-designed loss function part has also been updated.

Signed-off-by: Yiheng Wang <vennw@nvidia.com>

PR #131, opened 2021-02-26
Add new pipeline of DynUNet

yiheng-wang-nv:
Since the old tutorial notebook has many potential issues, recently I rewrite a new pipeline to place the old notebook. The work is still in process and I will submit a formal PR after finishing it.

To do list:

- inference script
- DDP support for large dataset, such as task 03, 07, 08 and 10 for decathlon.
- More doc strings.


yiheng-wang-nv:
This PR is closed in order to decouple with #126, I'll create a new branch for the new pipeline.

PR #138, opened 2021-03-13
BUG: MedNIST data partitioned differently for different seeds

adamaji:
Companion PR to https://github.com/Project-MONAI/MONAI/pull/1761 that fixes a similar issue. The data lists in this notebook are not partitioned in the same way depending on the random seed. Instead of using `np.random.random()` serially to select indices, choose all the indices at once based on one `np.random.shuffle`.


wyli (approved):
thanks!

PR #149, opened 2021-03-22
print number of successful tests

rijobro:
forgot to submit last commit earlier. prints number of successfully executed tests


wyli (approved):
thank you!

Issue #151, opened 2021-03-23
mednist_GAN_workflow_array has errors


wyli:
**Describe the bug**
full logs from https://github.com/Project-MONAI/MONAI/runs/2172199805?check_suite_focus=true

```
Running ./modules/mednist_GAN_workflow_array.ipynb
Checking PEP8 compliance...
Running notebook...
Before:
    "max_epochs = 50\n",
After:
    "max_epochs = 1\n",
Before:
    "disc_train_steps = 5\n",
After:
    "disc_train_steps = 1\n",

Executing:   0%|          | 0/33 [00:00<?, ?cell/s]
Executing:   3%|▎         | 1/33 [00:01<00:37,  1.19s/cell]
Executing:   9%|▉         | 3/33 [00:06<00:50,  1.67s/cell]
Executing:  12%|█▏        | 4/33 [00:09<00:53,  1.84s/cell]
Executing:  30%|███       | 10/33 [00:23<00:46,  2.03s/cell]
Executing:  52%|█████▏    | 17/33 [00:30<00:27,  1.70s/cell]
Executing:  61%|██████    | 20/33 [00:32<00:17,  1.38s/cell]
Executing:  79%|███████▉  | 26/33 [00:38<00:09,  1.30s/cell]
Executing:  85%|████████▍ | 28/33 [00:39<00:04,  1.06cell/s]
Executing:  85%|████████▍ | 28/33 [00:41<00:07,  1.47s/cell]
Traceback (most recent call last):
  File "/opt/conda/bin/papermill", line 8, in <module>
    sys.exit(papermill())
  File "/opt/conda/lib/python3.8/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/opt/conda/lib/python3.8/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/opt/conda/lib/python3.8/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/opt/conda/lib/python3.8/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/opt/conda/lib/python3.8/site-packages/click/decorators.py", line 21, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/opt/conda/lib/python3.8/site-packages/papermill/cli.py", line 250, in papermill
    execute_notebook(
[...]

PR #154, opened 2021-03-23
Fixes tutorial 151

wyli:
fixes https://github.com/Project-MONAI/tutorials/issues/151

this fixes the issue, but any idea which PR breaks the previous version? @Nic-Ma


rijobro (approved):

wyli:
> fixes #151
> 
> this fixes the issue, but any idea which PR breaks the previous version? @Nic-Ma

the difference comes from the recent update in MetricLogger https://github.com/Project-MONAI/MONAI/blob/83855b98f1c128f46b4db21a2f3ee55d0b5ffa5d/monai/handlers/metric_logger.py#L104

PR #205, opened 2021-04-29
Move the notebook to avoid runexample.sh error

YuanTingHsieh:
### Description
Need to do this otherwise the runexample.sh will try to use python to run this notebook directly

### Status
**Ready**

### Checks


wyli (approved):

rijobro:
Thanks. Presumably the more robust solution would be to change the `runexamples.sh` so that all entries like this:

```python
file in "3d_segmentation/ignite"/*train*
```

became this:

```python
file in "3d_segmentation/ignite"/*train*.py
```

PR #213, opened 2021-05-24
Video seg

rijobro:
### Description
Adds an initial video training notebook.

This tutorial is simplistic in that each of the frames is treated as being completely independent. At the end, the frames are compiled into videos to allow for video inference, see [here](https://github.com/rijobro/real_time_seg/) for an example of the inference executable (python and C++).

### Status
**Ready**

### Checks


binliunls:
Hi @rijobro , I got three suggestions and was not sure about if they are helpful. Maybe we can discuss about them.
1, In this part of the video dataset, I think you transfer a bgr image to rgb one implicitly. Could it be better if a parameter like "color_type", default as "rgb",  are given and used to decide whether the image channel should be changed.
image
2, In the "video_seg.ipynb" tutorial, would it be helpful if the training dataloader's "shuffle" parameter were set to "True"?
3, The timing part in "video_seg.ipynb" may not work when cuda device involved，since cuda is a asynchronous process.

rijobro:
@binliunls Changes have been incorporated.

Nic-Ma:
Depends on the Video dataset PR in MONAI.

Thanks.

binliunls (approved):

PR #214, opened 2021-05-25
Update inference example with latest API

Nic-Ma:
### Description
This PR updated the inference example to use latest API and added comments to explain.

### Status
**Ready**

### Checks


wyli (approved):
thanks!

PR #226, opened 2021-06-10
Fixes layerwise learning rate demo

wyli:
### Description
according to https://github.com/Project-MONAI/MONAI/pull/2356 and requires that PR

### Status
**Ready**

### Checks


Nic-Ma (approved):
I suggest to keep the print of the network structure to make users easily understand before running.
Others look good to me.

Thanks.

PR #296, opened 2021-08-01
Pathology Tumor Detection

drbeh:
### Description
This PR adds a pathology pipeline example to train and evaluate a model to detect tumors on whole-slide histopathology images. This pipeline uses ignite-based MONAI workflow.

### Status
**Ready**


drbeh:
@wyli, @Nic-Ma, it would be great if either of you can review and merge this PR. Thanks.

Nic-Ma:
Hi @drbeh ,

Thanks for the quick PR!
I think the `dataset_subset.json` is too long in the code base, could you please help put it in our google drive storage? And use MONAI download API to download it in this example?
@wyli , what do you think?

Thanks.

wyli:
yes, the current shared drive at https://drive.google.com/drive/u/2/folders/1ZOqwF44newvq-jOSgnbXRDWRD9DRq8hH

drbeh:
Sure, let me check that.

drbeh:
I moved the datasets to the google drive and updated the README with the download links. For testing this pipeline (without any performance gain), you should download two images (tumor_001 and tumor_101) as mentioned in the README from [here](https://drive.google.com/drive/folders/0BzsdkU4jWx9BUzVXeUg0dUNOR1U?resourcekey=0-dODmENBQPCw06DITRJfnfg)
You also need the `dataset_0_subset_0.json` from [here](https://drive.google.com/file/d/1NCd0y4FR42maQpfZjzKlFSIX4oeKgysg/view?usp=sharing).

wyli:
Looks good to me. My concern is that we'll probably break this tutorial when developing the core codebase. This happened before https://github.com/Project-MONAI/MONAI/issues/2571, and it was detected manually -- which is not efficient. 

So I'm wondering whether we should also convert this into an integration test in MONAI (e.g. https://github.com/Project-MONAI/MONAI/blob/dev/tests/test_integration_determinism.py), to download some testing data (e.g. https://github.com/Project-MONAI/MONAI/blob/7f42efe0ab55730c302aba87bcd9bf75c4bfc496/tests/test_patch_wsi_dataset.py#L27-L28), run a few training steps.

drbeh:
@wyli, I agree with you that we should have an integration test in MONAI for pathology too. I can make a simplified version of this for integration but it requires to download two large images (3.1GB and 1.4GB). Wouldn't it be a problem?

Anyways, We should have some examples in tutorials, shouldn't we?

wyli:
Adding tests will be great. gigabytes of testing data will cause issues, is it possible to have a smaller test?

Yes we should have examples in the tutorial, and ideally other developers should be able to help maintain it when you are not available.

wyli (commented):
thanks, created a separate ticket for the integration test https://github.com/Project-MONAI/MONAI/issues/2676. @Nic-Ma could you please help confirm/merge this PR?

Nic-Ma (approved):
Looks good to me.
Could you please help also update the README doc of this repo to include this pipeline?
https://github.com/Project-MONAI/tutorials/blob/master/README.md

Thanks in advance.

drbeh:
README is updated!

PR #301, opened 2021-08-04
299 Add kaggle ranzcr 4th place solution

### LINKED ISSUE #299: Reimplement 4th place solution of RANZCR challenge
Add the implementation in `tutorials/kaggle/RANZCR/4th_place_solution`, for the first step:
1. Replace the `CSV reading + OpenCV loading` logic with MONAI `CSVDataset + LoadImage transform`.
2. Replace all the albumentation transforms with MONAI transforms.

yiheng-wang-nv:
Fixes #299 .
The pipeline has been reproduced with MONAI and Pytorch APIs.
The scores on the testset for a single model (one seed and four seeds ensemble) are comparable with the original code:

late_sub_monai_reproduce

### Status
**Ready**

### Checks


kaggle/RANZCR/4th_place_solution/data/seg_data.py
```diff
@@ -0,0 +1,113 @@
+import ast
+
+import cv2
```
wyli, line 3: Is opencv installed by default? We may need so provide some install instructions

kaggle/RANZCR/4th_place_solution/models/seg_model.py
```diff
@@ -0,0 +1,246 @@
...
+                interp_mode=InterpolateMode.LINEAR,
+            )
+        if act is not None:
+            act_layer = get_act_layer(act)
+        else:
+            act_layer = nn.Identity()
+        super().__init__(conv_layer, up_layer, act_layer)
+
+
+class Net(nn.Module):
```
wyli, line 147: This should have a clear class name

kaggle/RANZCR/4th_place_solution/utils.py
```diff
@@ -0,0 +1,116 @@
...
+    return checkpoint
+
+
+def batch_to_device(batch, device):
+    batch_dict = {key: batch[key].to(device) for key in batch}
+    return batch_dict
+
+
+@jit
+def fast_auc(y_true, y_prob):
```
wyli, line 105: How is this different from monai's implementation?
Need to provide instructions about installing numba

kaggle/RANZCR/4th_place_solution/utils.py
```diff
@@ -0,0 +1,116 @@
...
+import torch
+from monai.optimizers.lr_scheduler import WarmupCosineSchedule
+from numba import jit
+from torch import optim
+from torch.utils.data import DataLoader, SequentialSampler
+
+from data.seg_data import CustomDataset
+
+
+def set_seed(seed=1234):
```
wyli, line 14: We should use the monai utility to do this, if needed, we enhance the monai utilities

kaggle/RANZCR/4th_place_solution/utils.py
```diff
@@ -0,0 +1,116 @@
...
+    if scheduler is not None:
+        checkpoint["scheduler"] = scheduler.state_dict()
+
+    if scaler is not None:
+        checkpoint["scaler"] = scaler.state_dict()
+    return checkpoint
+
+
+def batch_to_device(batch, device):
+    batch_dict = {key: batch[key].to(device) for key in batch}
```
wyli, line 100: Recently we added the todevice transform, it should be used if possible

kaggle/RANZCR/4th_place_solution/utils.py
yiheng-wang-nv: Hi @wyli , this part can be removed since only loss is used for training. Let me change the code for it.

kaggle/RANZCR/4th_place_solution/utils.py
yiheng-wang-nv: Hi @wyli , do you mean `set_determinism`? this function use fixed `torch.backends.cudnn.deterministic = True` and `torch.backends.cudnn.benchmark = False`, which are not equal to this pipeline.
Shall I add another `set_seed` function in monai utilities?

kaggle/RANZCR/4th_place_solution/utils.py
yiheng-wang-nv: Sure, let me replace it.

kaggle/RANZCR/4th_place_solution/utils.py
wyli: we could use `set_determinism(seed)` and then set `torch.backends.cudnn.deterministic = False`, `torch.backends.cudnn.benchmark = True`. but I don't understand how this achieve deterministic, maybe they prefer speed rather than determinism?

kaggle/RANZCR/4th_place_solution/utils.py
```diff
@@ -0,0 +1,116 @@
...
+from numba import jit
+from torch import optim
+from torch.utils.data import DataLoader, SequentialSampler
+
+from data.seg_data import CustomDataset
+
+
+def set_seed(seed=1234):
+    random.seed(seed)
+    os.environ["PYTHONHASHSEED"] = str(seed)
```
wyli, line 15: looks like this line of setting PYTHONHASHSEED has no effect, not sure whether we really need it ([ref](https://stackoverflow.com/questions/25684349/can-os-environpythonhashseed-be-set-dynamically-from-within-an-application))

kaggle/RANZCR/4th_place_solution/utils.py
wyli: sure, I think it's good to have todeviced in the training loop

wyli (approved):
looks good to me

PR #368, opened 2021-09-23
Ziyue Changes to Performance Notebooks

drbeh:
### Description
I am submitting this PR on behalf of Ziyue. This PR updates performance notebooks for both pathology and radiology.

### Status
**Ready**

### Checks


wyli:
I had another look at the files, 
- the notebook is quite large (1.1MB), perhaps the figures are already embedded in it, and there's no need to include a folder of `Figures`?  
- another thing is that do we really need the information being in the jupyter notebook format? to me `profiling_train_base_nvtx.ipynb` could be replaced by a simple markdown file without losing much practical value.

drbeh:
> I had another look at the files,
> 
> * the notebook is quite large (1.1MB), perhaps the figures are already embedded in it, and there's no need to include a folder of `Figures`?
> * another thing is that do we really need the information being in the jupyter notebook format? to me `profiling_train_base_nvtx.ipynb` could be replaced by a simple markdown file without losing much practical value.

Thanks @wyli! They are good points. I'll be out of office next week, so could you please sync with Daguang/Dong/Ziyue about these notebooks?

wyli (approved):

PR #388, opened 2021-10-05
update densenet_training_array.py

tommydino93:
removed redundant line

Fixes # .

### Description
The removed line is useless since there is already `best_metric = -1` which ensures we enter the if statement at least once

### Status
Ready


tommydino93:
Hello :)

Is it normal that the pep8 check is taking hours to complete?

wyli:
no it's not.. looks like the service is down at the moment https://www.githubstatus.com/incidents/bdbzpz7qxmbx

PR #487, opened 2021-12-16
[WIP] 486 Add example model package

### LINKED ISSUE #486: Add example model package
Sub task of ticket https://github.com/Project-MONAI/MONAI/issues/3482
Create an example of model package.

Nic-Ma:
Fixes #486 .

### Description
This PR implemented a draft example of the model package for discussion, leverage @ericspod 's great work to save data: https://github.com/Project-MONAI/MONAI/pull/3138

1. Basic principles
(1) **Structured sharable data**: Define meta data and components in structured files, like JSON or YAML. With predefined schema to verify the JSON configs, follow: https://json-schema.org/. We can put the base common schema in some public web storage in the future.
(2) **TorchScript package**: Export model weights, meta data, components config, etc. into TorchScript file.
(3) **Hybrid programming**: For 90% cases, only the JSON configs should be enough. For some advanced cases, the developer of model package defines which parts of the workflow should be shared for others to reconstruct other logic, then define these parts in **JSON or YAML config**, and can implement other customized logic in the **python program**. For example, this PR defines `transforms`, `dataset`, `dataloader`, `inferer`, etc. in the `inference.json` config, implements a task-specific special inference logic with native PyTorch program in `inference.py`, other teams can easily leverage the config file to reconstruct necessary components and implement their specific inference logic.
(4) **Data sharing**: the python program can leverage the `ConfigParser` to get constructed instance or raw config items from config file then **lazy instantiation**, configs can **refer / extend** other items in a file or even other files.

2. Structure of the package
(1) **commands**: start point to execute, like: `export.sh` to export the network weights, meta, config, etc. to TorchScript model, `inference.sh` to load TorchScript model, construct instances and execute inference. We should enhance it to something like `monairun` to easily support all platforms and environments.
(2) **configs**:  structure config for shareable components or information, can be JSON or YAML, like: `metadata.json` record the necessary meta information of the model package, `inference.json` define the args and components for inference usage.
(3) **docs**: necessary documents and images, etc. like: `README.md`, `lisence.txt`, `tensorboard.png`, etc. depends on the model context.
(4) **models**: pretrained model weights and exported TorchScript model file.
(5) **[Optional] scripts**: if the model package has customized python program, like `inference.py` leverage the structured components to construct a special inference program.

### Status
**Work in progress**

### Checks


Nic-Ma:
I didn't implement any config-parsing related logic in it so far, marked as `TODO:` in the code.

Nic-Ma:
Interesting comments from @ericspod during the online meeting:
1. Just a note that .ts is meant to be Torchscript but to web developers it'll looks like a Typescript file, we may want to choose something else if anyone finds that confusing.
2. Versioning the package would help, if we used git repos to store things the commit hash becomes part of the version implicitly.

Thanks.

Nic-Ma:
Hi @ericspod @wyli ,

Maybe I misunderstood your suggestion, how can we get the commit hash if we want to save the commit hash in meta data for this commit...?

Thanks in advance.

ericspod:
> Hi @ericspod @wyli ,
> 
> Maybe I misunderstood your suggestion, how can we get the commit hash if we want to save the commit hash in meta data for this commit...?
> 
> Thanks in advance.

What I was thinking was, from the perspective of a user, the commit hash could be used to determine model versions. It wouldn't be included in the package but would be used to tell model versions apart that claim to be the same version, ie. if the user forgets to increment the version.

Nic-Ma:
> > Hi @ericspod @wyli ,
> > Maybe I misunderstood your suggestion, how can we get the commit hash if we want to save the commit hash in meta data for this commit...?
> > Thanks in advance.
> 
> What I was thinking was, from the perspective of a user, the commit hash could be used to determine model versions. It wouldn't be included in the package but would be used to tell model versions apart that claim to be the same version, ie. if the user forgets to increment the version.

Thanks for your explanation, looks good to me!

modules/model_package/spleen_segmentation/programs/__init__.py
```diff
@@ -0,0 +1,11 @@
+
```
wyli, line 1: perhaps the name of this folder could be `scripts`

Nic-Ma: Good point, updated in the latest commit.

Thanks.

Nic-Ma:
Slightly adjusted the folder structurer, thanks for the reference of MONAI application:
https://github.com/Project-MONAI/monai-deploy/blob/main/guidelines/monai-application-package.md#table-of-important-paths

Thanks.

Nic-Ma:
Hi @wyli @ericspod ,

Thanks for several internal discussions, I updated the example model package to show new ideas:
1. `inference.py` shows how to get config and modify it then instantiation, and also shows how to get instance directly, these are the 2 ways for JSON / python hybrid.
2. `inference_v2.py` shows how to write a new config with a base config and override some configs.
3. `trtinfer.json` and `trtinfer.py` show how to refer to config in another JSON file or same file, simulated TensorRT and DALI inference logic.

Could you please help take a look again?

Thanks in advance.

Nic-Ma:
Hi @ericspod @wyli ,

I added the `tensorboard.png` and `mlflow.png` into the model package for training curves, and changed the `docs` folder to `materials`.

Thanks.

modules/model_package/spleen_segmentation/configs/metadata.json
```diff
@@ -0,0 +1,30 @@
...
+    "monai_version": "0.8.0",
+    "pytorch_version": "1.10.0",
+    "numpy_version": "1.21.2",
+    "optional_packages_version": {"nibabel": "3.2.1"},
+    "task": "Decathlon spleen segmentation",
+    "description": "A pre-trained model for volumetric (3D) segmentation of the spleen from CT image",
+    "authorship": "MONAI team",
+    "data_source": "Task09_Spleen.tar from http://medicaldecathlon.com/",
+    "dataset_dir": "/workspace/data/Task09_Spleen",
+    "image_classes": "single channel data, intensity scaled to [0, 1]",
```
ericspod, line 16: These tags are describing the input and output formats, that is shape and scaling. I would have this broken out into explicit tags so that this information can be used to construct the appropriate transforms to format inputs and posttransforms for outputs. Something like this adds more specific information:

```json
"data_format":{
    "inputs":[  
        {
            "name": "image",
            "type": "image",
            "format": "magniture",
            "shape": [1, 160, 160, 160],
            "type": "float32",
            "value_scale": [0,1],
            "center_crop": true,
        },
    ],
    "outputs":[
        {
            "name": "pred",
            "type": "image",
            "format": "segmentation",
            "from_network": {
                "shape": [2, 160, 160, 160],
                "type": "float32",
            },
            "shape": [1, 160, 160, 160],
            "type": "uint8",
            "classes": { 0: "background", 1: "spleen"},
            "argmax": true
        },
    ],
    "patch_inference": false
}
```

This provides information on what the input to the network is expected to be, that is a single float32 (1,160,160,160) image with values scaled to [0,1], so inferring what transforms to use to get this from arbitrary inputs can be done. For the output we have one image with an initial format having 2 channels which argmax is applied to to get the final 1 channel segmentation image, again using this information the transforms to do this can be inferred. 

Other information can be added to indicate how the network is to be applied to inputs, so "patch_inference" indicates the network is fed the full input image rather than being applied to patches of shape (1,160,160,160) from the input image.

modules/model_package/spleen_segmentation/configs/inference.json
```diff
@@ -0,0 +1,142 @@
...
+            "spatial_dims": 3,
+            "in_channels": 1,
+            "out_channels": 2,
+            "channels": [16, 32, 64, 128, 256],
+            "strides": [2, 2, 2, 2],
+            "num_res_units": 2,
+            "norm": "batch"
+        }
+    },
+    "preprocessing": [
```
ericspod, line 17: I commented below on defining the input and output formats more verbosely so that a transform sequence like this can be inferred with less JSON. This kind of definition is still needed for cases where that isn't going to have enough specific information, so we should establish the JSON format to use for any given transform. For transforms like "Spacingd" it's clear how to provide values in JSON with Python equivalents for each constructor argument, but more complex transforms may need more help. In general though we can represent instantiating a transform using this JSON syntax where a mapping `{"name": "Foo", "args":{...}}` is transformed into a constructor call for class `Foo`.

aihsani (commented):
Not sure why we need shell scripts in the model package... they're not easily portable across operating systems, and not descriptive in what they each achieve. Wouldn't in make more sense to remove any shell scripts and implement functionality in MONAI that is able to interpret/generate the JSON artifacts via a CLI based on the framework being used?

modules/model_package/spleen_segmentation/configs/inference.json
Nic-Ma: @ericspod thanks for the idea, and about the instantiation logic, I am already developing it in draft PR: https://github.com/Project-MONAI/MONAI/pull/3593/files#diff-7041497533453f2514ad444d001950097e020fce4b9c2e5afee00edb99707d64R72.

Thanks.

modules/model_package/spleen_segmentation/configs/metadata.json
Nic-Ma: Interesting idea! I will try to enhance the metadata JSON.
I think it's a long way to go to finalize a perfect metadata dictionary, so for the first version, we should try to keep it simple and easy to use. Then we can add more and more items to it based on users' feedback, and also develop the `schemes` to format expected content.

Thanks.

Nic-Ma:
Hi @aihsani ,

Thanks for your suggestion!
I think the `shell scripts` are used to simplify the common use cases and args, especially for researchers to tune args when training a MMAR, for example, they may have several different configs, change config file in shell.
For example, the `inference.sh` in this example is:
```
python ../scripts/inference.py
    --config ../configs/inference.json
    --meta ../configs/metadata.json
```
And you are right, we should definitely not put complicated logic in the `shell script`, main things should in the python or JSON files, you can call the python program `inference.py` directly.

Thanks.

modules/model_package/spleen_segmentation/configs/metadata.json
Nic-Ma: @ericspod , and Ming also provided more information in: https://github.com/Project-MONAI/MONAI/issues/3482#issuecomment-1018245931.

Thanks.

Nic-Ma: Hi @ericspod ,

Thanks for your suggestion, I updated the `metadata.json` in this PR according to your suggestion.
I slightly changed several things in your proposal:
1. Changed the section name from `data_format` to `network_data_format`, as it actually only defines the input data for the network, not the data source for transforms, Dataset or DataLoader, etc.
2. Changed the `list` in `inputs` and `outputs` to `dict`, I feel it's more clear to access the content by `keys`.
3. I didn't define the `postprocessing` logic (like `argmax`, etc.) in the first version, as we also don't define the information for `preprocessing` here. I agree we can consider to add them in future, let's try to keep the metadata simple for the first version?

Could you please help take a look again?

Thanks in advance.

modules/model_package/spleen_segmentation/configs/inference.json
Nic-Ma: Hi @ericspod ,

Updated the PR to refer to the `metadata` configs.
More details about how to parse the configs is in this draft PR: https://github.com/Project-MONAI/MONAI/pull/3593.

Thanks for your suggestion!

Nic-Ma:
I added `schema` example to check the `metadata` format and expected values.
It's based on the `jsonschema` standard: http://json-schema.org/.
I didn't define schema for all the items, just included key required items as example.

Thanks.

Nic-Ma:
Hi @wyli ,

I changed the `override` feature to a separate MMAR example according to our online discussion to make it more clear.

Thanks.

Nic-Ma:
Hi @ericspod @wyli @rijobro @aihsani ,

I have updated this MMAR example according to our online discussion yesterday.
And moved the `python scripts` and `schema` to MONAI core to support 90% cases:
https://github.com/Project-MONAI/MONAI/tree/506b9089637d7dd9ed3496dde9deee8c669f269d/monai/apps/mmars

Thanks.

ericspod:
To recap what we had discussed about actual file structure, let's say we have a directory with at least these files:

```
models/model.ts
models/model.pt
configs/metadata.json
```

We can consider this directory the model package itself. The metadata.json would have in it the information that is currently provided in this PR, plus it should also have the arguments to instantiate the network rather than that being in the inference config. If the model isn't compatible with Torchscript then the mode.ts file will be absent and this will be needed to recreate the network before loading the stored weights. 

This directory can then be packaged into a zip file whose structure can be assumed with `models` and `configs` directories. If the model is compatible with Torchscript then a new .ts file can be made with the metadata stored internally and other files saved as extra files using [save_net_with_metadata](https://github.com/Project-MONAI/MONAI/blob/7c9fefb55e13007e798f04bccc560718b3b03dbb/monai/data/torchscript_utils.py#L26). Users can then load this Torchscript model without worrying about the metadata or anything else MONAI specific if they just wanted to get things started quickly. Either way we'd have a zip file that's easy to distribute and which other software can expect to have as input.

wyli:
closing in favour of https://github.com/Project-MONAI/tutorials/tree/master/modules/bundles

PR #489, opened 2021-12-20
Add openfl example

alexey-khorkin:
### Description
Hi! We would like to add a federated learning example with [OpenFL](https://github.com/intel/openfl) framework. This example based on [registration_mednist.ipynb](https://github.com/Project-MONAI/tutorials/blob/master/2d_registration/registration_mednist.ipynb)


federated_learning/openfl/openfl_mednist_2d_registration/envoy/mednist_dataset_shard_descriptor.py
```diff
@@ -0,0 +1,109 @@
+# Copyright (C) 2020-2021 Intel Corporation
```
wyli, line 1: thanks, I have an impression that the licensing blob should follow the licensing informantion here https://github.com/Project-MONAI/MONAI/blame/dev/CONTRIBUTING.md#L66-L77, could you please revise?

alexey-khorkin: @wyli, thanks for comment, we think it is better to remove the license information since this is an example of user code by purpose

wyli (approved):
Thanks, it looks good to me.

PR #493, opened 2021-12-23
interpretability with cats and dogs

rijobro:
I was unsure of our implementations of GradCAM and occlusion sensitivity, so I tested on a binary dog/cat classification task.

I think these results are easier to understand than our other interpretability notebooks:

image

Requires PR https://github.com/Project-MONAI/MONAI/pull/3543.

### Status
**Ready**

### Checks


wyli:
perhaps exclude this file from the testing pipelines because the dataset is not automatically available https://github.com/Project-MONAI/tutorials/blob/1608664a2cb59b3043355fd4a48680c034c75247/runner.sh#L61-L66

rijobro:
@wyli , thanks done!

wyli (approved):

PR #495, opened 2021-12-29
update class_lung_lesion.ipynb

tommydino93:
Added replace=False in the explainability cause we don't want to investigate a sample more than once

Fixes # .

### Description
when choosing the samples to explain, we don't want to explain a sample more than once

### Status
**Ready/Work in progress/Hold**

### Checks


wyli (approved):
Thanks, it looks good to me.

PR #549, opened 2022-02-15
reduce duplication in 3d brats segmentation

rijobro:
As mentioned here, https://github.com/Project-MONAI/MONAI/discussions/3215 should only be accessing the validation dataset once so as to avoid repeatedly calling the transforms.

### Status
**Ready**

### Checks


wyli (approved):
Thanks, it looks good to me.

PR #553, opened 2022-02-16
Fix vector field plotting in DeepAtlas tutorial

ebrahimebrahim:
- When plotting a vector field as a quiver, the arrow x and y components
were swapped.
- When visualizing a displacement vector field via a deformed grid, the
inverse of the actual warping transformation was shown. This is because
cartesian coordinate paths were being displaced, rather than image
pixels.

The fixes are in the  plotting tools in `utils.py`, which are used to visualize results at the end of the tutorial.

Re-ran the tutorial notebook to show fixed output cells. Some differences in the output are due to running in a different environment. I also lowered the default image size from 128x128x128 to 96x96x96. This was just because of gpu memory limitations in the environment where I re-ran the notebook. Tutorial still looks fine at size 96.

### Status
Ready

### Checks

- [x] Notebook runs automatically `./runner.sh -p "-and -name '*deep_atlas*'"`


wyli (approved):
Thanks, it looks good to me.

PR #568, opened 2022-02-23
Update tutorials for 0.8.1

yiheng-wang-nv:
Signed-off-by: Yiheng Wang <vennw@nvidia.com>

Fixes #565 .

### Description
This PR fixes some issues in some tutorials for monai==0.8.1.

For `unet_segmentation_3d_ignite.ipynb`, the api needs to be updated.
For `threadbuffer_performance.ipynb`, the length of `strides` argument is not correct (has one extra value which will not be used).
For the multi-modal tutorial, `scikit-learn==0.20.3` is not compatible with python 3.8, it needs to be replaced with v 1.0.2 to solve the error.
For `cats_and_dogs.ipynb`, I added some doc-strings to explain about how to set the `directory`, and added the install of `ipywidgets` to prevent a related error when download the pretrained weights.

However, there are still some tutorials that are not covered or have issues:

1. `pathology/` (need @drbeh for help to check)
2. `self_supervised_pretraining/` (everything is good except the uploaded pretrained weights, see #575 )
3. `performance_profilling/` (also shown in https://github.com/Project-MONAI/tutorials/issues/453, @dongyang0122 helped to figure out that using `num_workers=0` can avoid the issue, however, I'm not sure if it is enough. May need some time to find the root cause.

### Status
**Ready**

### Checks

- [ ] Notebook runs automatically `./runner [-p <regex_pattern>]`

Issue #565, opened 2022-02-21
Verify tutorials with MONAI v0.8.1


Nic-Ma:
**Is your feature request related to a problem? Please describe.**
As we already released MONAI v0.8.1, it's better to double confirm the tutorials and examples with it.
Then we can tag v0.8.1 for tutorials repo.

yiheng-wang-nv:
Hi @Nic-Ma , I'm working on this

yiheng-wang-nv:
Hi @drbeh , could you please help to verify the tutorials in `pathology/`?
The dataset in https://www.kaggle.com/c/prostate-cancer-grade-assessment/ is >400 GB, and my hard disk cannot take it

drbeh:
Hi @yiheng-wang-nv, sure I will check the tutorial.

PR #573, opened 2022-02-23
Update Tumor Detection Profiling for v0.8.1

drbeh:
Partly addresses #565 

### Description
Update and verify tumor detection pipeline with MONAI v0.8.1 (https://github.com/Project-MONAI/tutorials/pull/568)

### Status
**Ready**

### Checks


yiheng-wang-nv:
Hi @drbeh , thanks for the updates. Could you try to avoid changing the outputs of cells and only change the code itself? Otherwise it is not convenient to check the differences.

drbeh:
> Hi @drbeh , thanks for the updates. Could you try to avoid changing the outputs of cells and only change the code itself? Otherwise it is not convenient to check the differences.

Right! GitHub doesn't have the best tools to diff jupyter notebooks. We shouldn't keep the output unchanged while they are depending on the changed code. Anyways, for the sake of reviewing this notebook, you just need to run it and see if it is working.

yiheng-wang-nv:
> We shouldn't keep the output unchanged while they are depending on the changed code

Hi @drbeh , you can check the commits history and most of the changes are only for the code itself. Iike what I did in #565

>  for the sake of reviewing this notebook, you just need to run it and see if it is working

If so, all reviewers need to run all notebooks every time to avoid checking thousands of diffs.

HI @wyli @Nic-Ma , could you please how to double confirm it? Thanks!

wyli:
Normally if there's no major change we often keep the previous output, it doesn't add much burden to the git history size and review becomes straightforward. That's really considerate. But I guess it's difficult to enforce this rule for every PR.

wyli (approved):
Thanks, it looks good to me.

drbeh:
> > We shouldn't keep the output unchanged while they are depending on the changed code
> 
> Hi @drbeh , you can check the commits history and most of the changes are only for the code itself. Iike what I did in #565
> 
> > for the sake of reviewing this notebook, you just need to run it and see if it is working
> 
> If so, all reviewers need to run all notebooks every time to avoid checking thousands of diffs.
> 
> HI @wyli @Nic-Ma , could you please how to double confirm it? Thanks!

@yiheng-wang-nv, I agree with you! What I typically do is to keep the outputs intact for each commit and re-run the affected outputs only at the last commit. In this way, we can compare to the one-to-the-last commit for checking the differences while using the last one to merging (assuming the last one only have changes in the outputs).
However, in this notebook, changing the code only and saving the notebook made so many changes in the metadata, which seems to be re-ordering of fields mainly. I think it was due to the difference in the version of jupyter engines.

Anyways, I agree that we should be more mindful of the changes in the notebooks for an easier review.

PR #649, opened 2022-04-07
648 Enhance the get_started tutorial of bundle

### LINKED ISSUE #648: Enhance MONAI Bundle tutorial
Here I wish to provide some suggestions about MONAI Bundle tutorial: see [here](https://github.com/Project-MONAI/tutorials/blob/a810488d96b878f830473fb5b6498b035f010617/modules/bundles/get_started.ipynb) 

1. In the section of `Execute training with bundle script - run`, the cells cannot be run in the notebook directly, thus should we change them into markdown cells?
2. As for the override function, it is a very helpful feature, but we may need to add this part in the doc: https://docs.monai.io/en/latest/bundle_intro.html (it seems missing). In addition, the sign `\` confuses me since it is not existing in the doc-strings of `def run` in the source code.

Nic-Ma:
Fixes #648 .

### Description
This PR enhanced the get_started tutorial of MONAI bundle according to the feedback.

### Status
**Ready**

### Checks

### DIFF
--- a/modules/bundles/get_started.ipynb
+++ b/modules/bundles/get_started.ipynb
@@ -517,12 +517,10 @@
    ]
   },
   {
-   "cell_type": "code",
-   "execution_count": null,
+   "cell_type": "markdown",
    "metadata": {},
-   "outputs": [],
    "source": [
-    "%python -m monai.bundle run \"'train#trainer'\" --config_file configs/train.json"
+    "`python -m monai.bundle run \"'train#trainer'\" --config_file configs/train.json`"
    ]
   },
   {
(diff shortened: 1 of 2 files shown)


modules/bundles/get_started.ipynb
```diff
@@ -517,12 +517,10 @@
    ]
```
yiheng-wang-nv, line 517: <p>Here I suggest to add:</p><p><code></code><code></p><p>If the escape character <code>\</code> is needed in your shell, <code>--device "\$torch.device('cuda:1')"</code> should be used. </p><p></code><code></code></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Project-MONAI/tutorials/pull/649/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='33'/>

Nic-Ma: updated according to your comments.

Thanks.

yiheng-wang-nv (approved):
Thanks Nic!

PR #653, opened 2022-04-11
652 Enhance `get_kernels_strides` of DynUNet pipeline

yiheng-wang-nv:
Fixes #653.

### Description
Add some doc-strings to describe the requirements of the input size for DynUNet, and for sizes that cannot find suitable strides, an error will be raised. This change is helpful for users that need to use this function for other tasks.

### Status
**Ready**

### Checks

### DIFF
--- a/modules/dynunet_pipeline/create_network.py
+++ b/modules/dynunet_pipeline/create_network.py
@@ -32,7 +32,7 @@ def get_kernels_strides(task_id):
         for idx, (i, j) in enumerate(zip(sizes, stride)):
             if i % j != 0:
                 raise ValueError(
-                    f"patch size error, please try to modify the size {input_size[idx]} in dimension {idx}."
+                    f"patch size error, please try to modify the size {input_size[idx]} in the spatial dimension {idx}."
                 )
         sizes = [i / j for i, j in zip(sizes, stride)]
         spacings = [i * j for i, j in zip(spacings, stride)]
--- a/modules/dynunet_pipeline/create_network.py
+++ b/modules/dynunet_pipeline/create_network.py
@@ -7,22 +7,38 @@ from task_params import deep_supr_num, patch_size, spacing
 
 
 def get_kernels_strides(task_id):
-    sizes, spacings = patch_size[task_id], spacing[task_id]
-    strides, kernels = [], []
+    """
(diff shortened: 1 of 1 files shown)


modules/dynunet_pipeline/create_network.py
```diff
@@ -7,22 +7,38 @@
...
+    be divisible by the product of all strides in the corresponding dimension.
+    In addition, the minimal spatial size should have at least one dimension that has twice the size of
+    the product of all strides. For patch sizes that cannot find suitable strides, an error will be raised.
+
+    """
     sizes, spacings = patch_size[task_id], spacing[task_id]
+    input_size = sizes
     strides, kernels = [], []
-
+    min_patch_size = [1, 1, 1] * len(sizes)
```
ericspod, line 21: This variable doesn't appear used anywhere except to update itself?

yiheng-wang-nv: Thanks Eric, I forgot to remove it.

modules/dynunet_pipeline/create_network.py
```diff
@@ -7,22 +7,38 @@
...
             for (ratio, size) in zip(spacing_ratio, sizes)
         ]
+        min_patch_size = [i * j for i, j in zip(min_patch_size, stride)]
         kernel = [3 if ratio <= 2 else 1 for ratio in spacing_ratio]
         if all(s == 1 for s in stride):
             break
+        for idx, (i, j) in enumerate(zip(sizes, stride)):
+            if i % j != 0:
+                raise ValueError(
+                    f"patch size error, please try to modify the size {input_size[idx]} in the spatial dimension {idx}."
```
ericspod: This would be better if it said something specific about the condition being checked that helped the user choose better values. I'm not totally sure that's the condition that's being checked here but it would help more to say something about what is checked and how to fix it, perhaps something like f"Patch size error, please try to modify the size {input_size[idx]} in the spatial dimension {idx} to be an even multiple of the patch size."

ericspod: You updated the code as I wrote this! The new message looks better, I suppose it's too complex to say anything more specific about what patch size is supported.

ericspod (approved):

PR #680, opened 2022-05-02
initial reference implementation for Breast Density FL Challenge

holgerroth:
Fixes # n/a.

### Description
Reference implementation for 
[ACR-NVIDIA-NCI Breast Density FL challenge](http://BreastDensityFL.acr.org).

Held in conjunction with [MICCAI 2022](https://conferences.miccai.org/2022/en/).

We provide a minimal example of how to implement Federated Averaging using [NVFlare 2.0](https://github.com/NVIDIA/NVFlare) and [MONAI](https://monai.io/) to train
 a breast density prediction model with ResNet18.

### Status
**Ready**

### Checks
n/a


Nic-Ma:
I feel the JSON dataset files are big, maybe better to put them in our Google Drive?
@wyli What do you think?

Thanks.

holgerroth:
> I feel the JSON dataset files are big, maybe better to put them in our Google Drive? @wyli What do you think?
> 
> Thanks.

I could compress them and ask user to unzip.

federated_learning/breast_density_challenge/Dockerfile
```diff
@@ -0,0 +1,36 @@
+# use python base image
```
wyli, line 1: this doesn't seem to include a cuda/pytorch base, is it expected?

federated_learning/breast_density_challenge/README.md
```diff
@@ -0,0 +1,176 @@
...
+
+Note, the location of the dataset and data lists will be given by the system.
+Do not change the locations given in [config_fed_client.json](./code/configs/mammo_fedavg/config/config_fed_client.json):
+```
+  "DATASET_ROOT": "/data/preprocessed",
+  "DATALIST_PREFIX": "/data/dataset_blinded_",
+```
+
+### 1.2 Build container
+The argument specifies the FQDN of the FL server. Use `localhost` when simulating FL on your machine.
```
wyli, line 36: perhaps define `FQDN`?

federated_learning/breast_density_challenge/Dockerfile
holgerroth: We used the nvidia pytorch base image first but the container image size ended up being >15 GB. Same with the MONAI base image. Using just the Python base image like this, the container size is reduced to just 4 GB. That's desireable for the challenge platform.

federated_learning/breast_density_challenge/README.md
holgerroth: Good point. Added the definition.

PR #694, opened 2022-05-10
Add DeepEdit Bundle

diazandr3s:
Signed-off-by: Andres Diaz-Pinto <diazandr3s@gmail.com>

DeepEdit Bundle

PR #710, opened 2022-05-16
Update spleen bundle config as example

Nic-Ma:
### Description
This PR updated the spleen bundle config according to the discussion: https://github.com/Project-MONAI/tutorials/pull/694#discussion_r872406004

### Status
**Ready**

### Checks


ericspod (approved):
Looks good, I think this is a much cleaner style to be using in these specifications.

PR #728, opened 2022-05-23
Update Multiple Instance Learning Pipeline

### LINKED ISSUE #637: [MONAI Pathology] Update multiple instance learning pipeline
There are going to be major changes in pathology components in MONAI, so the multiple instance learning pipeline needs to be updated with these new components.

drbeh:
Fixes #637 

### Description
This PR updates Multiple Instance Learning Pipeline with the new components in MONAI.

### Status
**Ready**

### Checks


drbeh:
@Nic-Ma Here is the accuracy metric of the updated pipeline with respect to the original one that Andriy has run the experiment and produced the figures. The dark blue is the original one and the orange is the updated one (ignore the other colors). Both dark blue and orange ones are performing very similarly considering the different randomness that they have.
Screen Shot 2022-06-02 at 11 05 51 AM

drbeh:
@Nic-Ma I believe this pipeline is already tested by @myron and equivalent performance is reported. Can you please help to merge it?

pathology/multiple_instance_learning/panda_mil_train_evaluate_pytorch_gpu.py
```diff
@@ -259,7 +207,7 @@ class LabelEncodeIntegerGraded(Transform):
 
     """
 
-    def __init__(self, num_classes, keys=["label"]):
+    def __init__(self, num_classes, keys=("label",)):
```
Nic-Ma, line 210: If the transform works for `dict` data, could you please help change it to inherit from `MapTransform` base class?

Thanks.

pathology/multiple_instance_learning/panda_mil_train_evaluate_pytorch_gpu.py
```diff
@@ -540,5 +464,85 @@ def main_worker(gpu, args):
     print("ALL DONE")
 
 
+def parse_args():
```
Nic-Ma, line 467: I think maybe it's not necessary to change the function position?

Thanks.

drbeh: I think for the overall readability, it's better this way. Thanks

pathology/multiple_instance_learning/panda_mil_train_evaluate_pytorch_gpu.py
drbeh: Sure, will do that.

Nic-Ma (approved):
Thanks for the update, looks good to me.
@wyli Do you have any other comments?

Thanks.

PR #732, opened 2022-05-25
Dice tutorial

ericspod:
### Description
From discussions I put together an extended explanation of the dice loss and metric with examples.

### Status
**Ready**

### Checks


wyli (approved):
Thanks!

PR #734, opened 2022-05-26
733 Fix bundle metadata errors

### LINKED ISSUE #733: Spleen bundle `spatial_shape` not match in `metadata.json`
In https://github.com/Project-MONAI/tutorials/blob/32dfc1cf45f0fceeab4ab9e2828d571e317d8a1a/modules/bundle/spleen_segmentation/configs/metadata.json#L37
The `spatial_shape` is (160, 160, 160), but the spatial size for training and the roi size for inference is (96, 96, 96) ([see here](https://github.com/Project-MONAI/tutorials/blob/32dfc1cf45f0fceeab4ab9e2828d571e317d8a1a/modules/bundle/spleen_segmentation/configs/train.json#L77)).
The size should be unified.
In addition, I still have the following questions about the `metadata.json` part.

1. The `data_type` is "dicom" ([see here](https://github.com/Project-MONAI/tutorials/blob/32dfc1cf45f0fceeab4ab9e2828d571e317d8a1a/modules/bundle/spleen_segmentation/configs/metadata.json#L19)), but the data is in `nii.gz` form. Is it an error here? How should we define the `data_type` here?
2. How should we define the `format` of image ([see here](https://github.com/Project-MONAI/tutorials/blob/32dfc1cf45f0fceeab4ab9e2828d571e317d8a1a/modules/bundle/spleen_segmentation/configs/metadata.json#L35)). For spleen dataset, it is "magnitude", how about the brats dataset?
3. As for the `spatial_shape` mentioned above, if the input size of the network in training is different from the roi size used for inference, which one should be used for `spatial_shape` in the metadata? As for a segmentation model, different sizes are supported usually, thus I'm not sure how to define it.

Hi @ericspod , may need your help to explain these questions, thanks!

yiheng-wang-nv:
Signed-off-by: Yiheng Wang <vennw@nvidia.com>

Fixes #733 .

### Description
This PR is used to modify the false descriptions in `metadata.json` of `spleen_segmentation` bundle.

### Status
**Ready**

### Checks

### DIFF
--- a/modules/bundle/spleen_segmentation/configs/metadata.json
+++ b/modules/bundle/spleen_segmentation/configs/metadata.json
@@ -16,7 +16,7 @@
     "authors": "MONAI team",
     "copyright": "Copyright (c) MONAI Consortium",
     "data_source": "Task09_Spleen.tar from http://medicaldecathlon.com/",
-    "data_type": "dicom",
+    "data_type": "nibabel",
     "image_classes": "single channel data, intensity scaled to [0, 1]",
     "label_classes": "single channel data, 1 is spleen, 0 is everything else",
     "pred_classes": "2 channels OneHot data, channel 1 is spleen, channel 0 is background",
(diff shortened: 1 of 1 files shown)


ericspod (approved):
Looks correct to me, thanks.

PR #735, opened 2022-05-26
Revert the modifications of outputs in "fix bundle metadata errors (#734)"

yiheng-wang-nv:
Reverts Project-MONAI/tutorials#734

The format of `outputs` should be `segmentation`, and should not have `modility`.


ericspod (approved):
I missed that too.

Issue #169, opened 2021-04-01
deepgrow files not present


rijobro:
Deepgrow tutorials need files that are not present to run, see here: https://github.com/Project-MONAI/MONAI/runs/2243105725?check_suite_focus=true#step:6:298.

rijobro:
@SachidanandAlle Do you think you could have a look at this?

wyli:
not sure we'll get this fixed. I'm trying to use the option of `runner.sh -p` to exclude the deepgrow files from the tests. but not sure how to set the arg. @rijobro any idea?  otherwise let's move the tutorials out of the repo because we couldn't maintain those.

SachidanandAlle:
the deepgrow inference notebook needs pretrained models (for 2D and 3D) and sample image..
if there is some place in aws or so.. if we can push them, then we can download and access it as part of notebooks run.

https://drive.google.com/drive/folders/1lypRGUoybX5detODnt2QnWC_j_Xkw2ys?usp=sharing
Here we need deepgrow_2d.ts, deepgrow_3d.ts and _image.nii.gz) to be available through wget kind to access them in notebooks (instead of g-drive)

wyli:
ok these files are useful, monai provides function to download from gdrive

rijobro:
I agree that the easiest thing to do is simply to download them from gdrive provided they won't be deleted any time soon.

rijobro:
https://github.com/Project-MONAI/MONAI/blob/master/monai/apps/utils.py#L74-L112

wyli:
```py
from monai.apps import download_url

resource = "https://drive.google.com/uc?export=download&id=1xRHGm9kUvRLEWeWeGj9EmYc9EEy77cEg"
dst = os.path.join("..", "bentoml", "classifier.zip")
if not os.path.exists(dst):
    download_url(resource, dst)
```

YuanTingHsieh:
Should be fixed by the above PR.

YuanTingHsieh:
Actually two questions here:

1. I don't have ``%pip install gdown`` in the top. Do we want to add that?
2. Do we want to include the output (those graphs and execution result of each cell) into the notebook?

PR #757, opened 2022-06-16
Update the fast training with profiling tool

### LINKED ISSUE #729: adding a profiling flag
the fast training tutorial shows good results but doesn't show the useful tools/utilities that helped reaching this performance as well as the perf tuning process https://github.com/Project-MONAI/tutorials/blob/main/acceleration/fast_training_tutorial.ipynb

provide a `PROFILING` flag, when it's set to `True`, enable the relevant nvtx utils used in https://github.com/Project-MONAI/tutorials/tree/main/performance_profiling and reduce the number of total epochs (for quick code protyping), provide some descriptions on how to tune the values in practice.

read through https://github.com/Project-MONAI/tutorials/tree/main/performance_profiling/radiology
but it's only a simplified demo.

**Additional context**
https://github.com/Project-MONAI/MONAI/blob/dev/monai/utils/nvtx.py

yuchen-xu:
Signed-off-by: Yuchen Xu <yuchenx@nvidia.com>

Fixes #729  .

### Description
Made changes in acceleration/fast_training_tutorial.ipynb and added a profiling option. When not profiling, the notebook runs like before. When profiling, the notebook must be run in the terminal as one piece (command provided), as Nsight systems does not support running from inside the script.

### Status
**Ready**


Nic-Ma:
I think the outputs folder can be removed?
@wyli Could you please help take a look at this PR first? You raised the feature request.

Thanks in advance.

wyli:
thanks, it looks great. cc @AHarouni

some minor points perhaps beyond this profiling example:
- any insight about further improving the speed or GPU utilization?
- could try this new flag for thread loader? https://github.com/ericspod/MONAI/blob/22924f5709e8c3b7786f3e201bf91be24a44da6e/monai/data/thread_buffer.py#L182
- might be interesting to try this recent feature of instancenorm https://github.com/Project-MONAI/MONAI/blob/812275ae5f4a1bc8a69b70a9823632b6f30a5724/monai/networks/layers/factories.py#L253 the current network is with batch norm, but I think

yuchen-xu:
@Nic-Ma I think the outputs/ folder makes outputs easier to manage (so that they don't live in the same directory as all the other tutorials), and allows users to customize where they want the outputs to go. Open to more discussion/opinions.

yuchen-xu:
@wyli Thanks for the comments!

- Could you clarify a little further what kinds of insights you are looking for?
- The Thread Worker flag seems to be in development and not part of MONAI yet?
- It would be interesting to try InstanceNorm indeed, but as you noted, they go beyond the profiling example - shall we open a new issue along the lines of "integrating latest features into fast training tutorial"?

yuchen-xu:
Hi @wyli 

- The use_thread_workers argument does not seem to be available in the MONAI v0.9.0rc2 docker (see attached image). Also, in the fast version, the main bottleneck seems to be CacheDataset (~40 seconds) rather than ThreadDataLoader (~0.002 seconds). 
- I was able to replace batch norm with instance norm and instance_nvfuser (the one you cited above). In the first few epochs, instance norm does seem to be faster than batch norm (~10% improvement from 10 seconds), and instance_nvfuser a little faster than that, but the two instance norms result in a worse metric. I will perform some more testing and include that in the updated version.

Nic-Ma:
Hi @yuchen-xu ,

1. Please try with MONAI v0.9.0 docker instead of RC2. And no need to worry about the `preparing cache` time.
2. If `instance_nvfuser` doesn't help, no need to update the tutorial for it. Please do some more experiments to confirm.

Thanks in advance.

yuchen-xu:
@wyli @Nic-Ma I performed experiments on the two changes, here are the findings:

1. Thread Workers didn't seem to improve the speed of the tutorial. In the figure, "orig" refers to the original setup (`use_thread_workers=False`), "thread1/2/3/4" refers to `use_thread_workers=True, num_workers=1/2/3/4`. Clearly, "orig" is the fastest among these. This can be further verified by the second part of the image (tlpo = training load per operation), which is the average time it takes the `next` operation when called on an iterator on the train loader. Again, "orig" is fastest. Curiously, `num_workers=1` makes it worse too.
image

2. Keeping `use_thread_workers=False`, it seems that replacing Batch Norm with either Instance Norm or Instance_NVFuser makes it faster. Performance on loss and Dice metric ends up around the same level, although the trends are different; see figures below. Also, Instance_NVFuser has a significantly slow first operation (taking ~10s under the fast regime), but is faster on later operations. I'm using instance_nvfuser, but will note the option to use instance norm too.
fig1
fig2

Each experiment is run with 600 epochs and repeated 3 times to control variance.

The next version (which I expect to push tomorrow) will also fix a minor issue in the generated graph of Dice metrics, such that the Dice and loss graphs will both show epochs going to the same number (right now the Dice graph has wrong values on the x-axis).

wyli:
Nice analysis, thanks @yuchen-xu!

Nic-Ma (approved):
Thanks for the enhancement, it overall looks good to me.
Just some minor comments inline.

Thanks.

Nic-Ma:
Hi @yuchen-xu ,

I think maybe there is something wrong during your last round training.
The epoch time curve of **regular training** looks strange, it jumped from 60s to 40s:
image
And the fast training is much slower to achieve mean_dice=0.94 than your previous round training, because it uses much more epochs (105 vs 60), so maybe we should not use instance norm?
image

Thanks.

yuchen-xu:
@Nic-Ma Thanks for the review!
The regular epoch time is a little strange indeed, and I could see that happen if someone else is also using the GPU. I'll try to run it again.
As noted in an earlier post, instance_nvfuser norm has overall the fastest runtime, but is pretty close to batch norm in terms of metric trends. On the other hand, instance norm is slightly slower (but still faster than batch norm) and may have better metric trends. I'll try again with instance norm and see what we get.

acceleration/fast_training_tutorial.ipynb
```diff
@@ -308,24 +480,32 @@
...
     "4. `ThreadDataLoader`: uses multi-threads instead of multi-processing, faster than `DataLoader` in light-weight task as we already cached the results of most computation.\n",
     "5. `DiceCE` loss function: computes Dice loss and Cross Entropy Loss, returns the weighted sum of these two losses.\n",
-    "6. Analyzed the training curve and tuned algorithm: Use `SGD` optimizer, different network parameters, etc."
+    "6. Analyzed the training curve and tuned algorithm: Use `SGD` optimizer, different network parameters, etc.\n",
+    "7. `Instance_NVFuser` norm (`Norm.INSTANCE_NVFUSER`) instead of batch norm (`Norm.BATCH`). This norm may take a while to set up in the beginning, but it still results in faster training with comparable results. As an alternative, use instance norm (`Norm.INSTANCE`), which is slightly slower but has more uniform time cost. For more details, refer to the [source code](https://github.com/Project-MONAI/MONAI/blob/812275ae5f4a1bc8a69b70a9823632b6f30a5724/monai/networks/layers/factories.py#L253) of all defined norms.\n",
```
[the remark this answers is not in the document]
yuchen-xu: On second thought, I wonder if it's better to direct users to the `[API reference](https://docs.monai.io/en/stable/networks.html?highlight=instance%20norm#module-monai.networks.layers.Norm)`. The pro is that a new version will not invalidate the source code link, and the con is that the API reference doesn't have a direct link to the source code.

yuchen-xu:
As it turns out, instance_nvfuser norm takes 105 epochs/258 secs to achieve the target Dice of 0.94, compared to 95 epochs/144 secs for instance norm and 60 epochs/106 secs for batch norm (original). While the two instance norms result in faster per-epoch time, they take significantly longer to achieve our goal metric. Therefore, we will continue using batch norm in the tutorial. @wyli

wyli:
the notebook testing script will modify `max_epochs` variable to speed up the tests https://github.com/Project-MONAI/tutorials/blob/2da31b6bfdfe8d341955c4b39fe5ab3cb2b0d025/runner.sh#L84-L85
looks like this PR is not compatible with that. please help confirm

Nic-Ma (approved):
Thanks for the quick update.
Looks good to me now, @wyli do you have any other comments?

Thanks.

acceleration/fast_training_tutorial.ipynb
```diff
@@ -35,21 +35,30 @@
    ]
```
wyli, line 35: <pre class="ql-syntax" spellcheck="false">Line #169.                        for val_step in range(len(val_loader)):
</pre><p>this is causing an error:</p><p> </p><pre class="ql-syntax" spellcheck="false">
[2022-07-10T01:34:43.235Z] Running ./acceleration/fast_training_tutorial.ipynb
[2022-07-10T01:34:43.235Z] Checking PEP8 compliance...
[2022-07-10T01:34:43.235Z] stdin:488:25: B007 Loop control variable 'val_step' not used within the loop body. If this is intended, start the name with an underscore.
[2022-07-10T01:34:43.235Z]                     for val_step in range(len(val_loader)):
[2022-07-10T01:34:43.235Z]                         ^
[2022-07-10T01:34:43.235Z] Error: Try running with autofixes: --autofix.
[2022-07-10T01:34:43.235Z] 
</pre><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Project-MONAI/tutorials/pull/757/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='21'/>

yuchen-xu: That's bizarre, the runner.sh test passed fine on my end. Changing the line to the following should fix it:
`for _ in range(len(val_loader)):`

yuchen-xu: I will fix this in PR #786. Thanks.

wyli: I have added the patch as you suggested, it works now, thanks!

PR #807, opened 2022-07-19
A tutorial to use randomised permutation during training

vikashg:
### Description
While training, if a monai transform chain is defined. It so happens that all the images go through the same transformations. All the transformations are vector quantities, so switching the order of transforms will generate a new augmentation. In the present implementation if you apply any transform more than once, you need to write it multiple times. Instead we are proposing is that the users can specify a dictionary of name of transformation (like rotate, flip or translate etc) and the frequency of transformation (an integer like 1, 2, ). The Augmentation class will then generate random transformation chain on the fly during training. One of the cons of this approach is that the cache method probably cannot be used as the transformations are being generated at random

### Status
**Ready**

### Checks

- [x] A tutorial is created where the usage can be tested.


wyli (approved):
thanks, it looks good to me, I'll move it to the modules subfolder and update runner.sh.

vikashg:
Thanks @wyli

PR #838, opened 2022-07-31
initial unet recon demo commit

mersad95zd:
Signed-off-by: mersad95zd <m.zalbagi@gmail.com>

Fixes # .

### Description
This PR contains a demo to perform training/validation with U-Net for accelerated MRI reconstruction.

### Status
**Ready**

### Checks


Can-Zhao:
I was wondering if you could:
- Add workflow figure to explain the recon algorithm. Current one is too simple.
- Add training/validation curve, so the users can know whether their performance is normal or not.
- Report inference result in readme, and compare with leaderboard to show that this implementation gives reasonable result.

Thank you!

Can-Zhao:
Thanks. 
Could you remove the model pt? We will put it in model-zoo in the future, not in the tutorial.
I was not sure whether this golfer should be called 2d_reconstruction, since it works on the 2D slices of 3D volumes. Would it be better to put them in folder `reconstruction/MRI_reconstruction/`?

Can-Zhao:
Looks good to me. @wyli Could you help final check it?

wyli (approved):

PR #841, opened 2022-08-01
Concatenate and pass patch location in MIL pipeline

### LINKED ISSUE #840: MIL pipeline patch information in data loader
As reported by a user (https://github.com/Project-MONAI/MONAI/discussions/4794#discussioncomment-3296673), the multiple instance learning pipeline does not provide patch level information in the data loader, so it hinders any post-processing that relies on patch locations within the whole slide image.

drbeh:
Fixes #840 

### Status
**Ready**

### Checks


pathology/multiple_instance_learning/panda_mil_train_evaluate_pytorch_gpu.py
```diff
@@ -237,6 +237,7 @@ def list_data_collate(batch: collections.abc.Sequence):
     for i, item in enumerate(batch):
         data = item[0]
         data["image"] = torch.stack([ix["image"] for ix in item], dim=0)
+        data["patch_location"] = torch.stack([ix["patch_location"] for ix in item], dim=0)
         batch[i] = data
```
KumoLiu, line 241: l'm not sure should we also concatenate patch_size and offset to make the process more consistent?

drbeh: That's a good point, @KumoLiu, but patch_size and offset are identical for all the patches in a batch, so we need to only keep one of them instead creating an array of them like patch location.

KumoLiu (approved):

PR #879, opened 2022-08-24
Update brats_segmentation_3d.ipynb

ajulyav:
Convert to np array to tensor in transformers

Fixes # .

### Description
To proceed with ConvertToMultiChannelBasedOnBratsClassesd, data must be Tensor. Though, with no conversion to Tensor (fixed now), it is np. array and causes errors. 

### Status
**Ready**

### Checks


ericspod:
I would suggest using `EnsureTyped` instead, it'll call the same code in the end but it's the class we want to encourage using.

ajulyav:
> I would suggest using `EnsureTyped` instead, it'll call the same code in the end but it's the class we want to encourage using.

If so, this is also a possible solution to work with Tensors.

ajulyav (commented):
no errors

wyli (approved):

PR #889, opened 2022-08-30
updated to reflect that we only support SEG for now (not RTSTRUCT)

kirbyju:
### Description
updated to reflect that we only support SEG for now (not RTSTRUCT)


kirbyju:
cc @wyli - Let me know if I need to do anything else

wyli (approved):
Thanks, it looks good to me.

PR #904, opened 2022-09-07
VarNet demo

mersad95zd:
Fixes # .

### Description
Tutorial for using e2e-VarNet for MRI reconstruction.

### Checks


Can-Zhao:
Could you remove the redundant files like fastmri_ssim.py? You could provide a link in the tutorial readme.

Can-Zhao:
1) Please put data_split file in json format, and send it to Wenqi. He will upload it to some link. In your readme, you just need to add the download link.
2) In Dataset section, could you explain the data path in the folder, e.g., "download the file XXX to `./data`"?
3) Let U-net demo and Varnet-demo share the same Dataset section, e.g. "The data preparation is explained in ../[unet_demo](../unet_demo)"

Can-Zhao:
I just realized your U-net demo does not use train/val/test split, but instead teast on validation set. Could you change it to train/val/test as in varnet demo? Please them share the same data split so the users can compare the performance better. Thanks!

wyli (approved):

PR #976, opened 2022-10-07
Dataloaders

aasthajh:
Fixes # .

### Description
This is an example highlighting the use of custom dataloaders for multi channel images. Multi channel images are very commonly used in healthcare, this example shows one such use case in microscopy. 

### Checks

- [x] Data must be downloaded prior to running the notebook
- [x] Path to the parent data folder must be provided


wyli:
thanks, I think it's a good PyTorch example but it's not really using monai. (any suggestions @Nic-Ma @rijobro @ericspod?)

aasthajh:
Would it help to use a MONAI model instead of pytorch ResNext? I had used MONAI ViT with the same dataloaders so I can swap out the model if that helps

ericspod:
I think the key contribution here is the dataset class demonstrating how to combine multichannel images. This can be recast using MONAI components (`LoadImage`) to load files instead of using PIL directly and use our models and other components in the training process which is pretty straight forward. You could use our engine classes as well instead of the regular training loop if you wanted to reduce total code in the notebook.

aasthajh:
Thanks Eric, I can replace the non-MONAI components with the MONAI equivalent. Could you please direct me to the _engine classes_ you referred to and any examples for using the same?

ericspod:
> Thanks Eric, I can replace the non-MONAI components with the MONAI equivalent. Could you please direct me to the _engine classes_ you referred to and any examples for using the same?

The classes are in the `monai.engine` module. This tutorial and others in the Tutorials repo and in the Model Zoo help explain their use: https://github.com/Project-MONAI/tutorials/blob/main/acceleration/distributed_training/unet_training_workflows.py

nvahmadi:
I just saw that this PR is still open. It should be noted that in the meantime I worked with @aasthajh to create a multi-channel microscopy tutorial (already merged, see [notebook](https://github.com/Project-MONAI/tutorials/blob/main/microscopy/multichannel_microscopy_classification.ipynb)). This tutorial covers the topics proposed in this notebook PR, but is more MONAI native (e.g. using `ConcatItemsd` transform and `DenseNet169` network). Perhaps it makes sense to close this PR, since it is not really necessary anymore? What do you think @aasthajh?

PR #999, opened 2022-10-19
HoVerNet training pipeline

KumoLiu:
Fixes https://github.com/Project-MONAI/MONAI/issues/4878

### Description
HoVerNet training tutorial
- [x] ignite version pipeline

### Checks

- [ ] Notebook runs automatically `./runner [-p <regex_pattern>]`

PR #1027, opened 2022-11-08
HoVerNet tutorial

KumoLiu:
Fixes [#4882.](https://github.com/Project-MONAI/MONAI/issues/4882)

### Description
A torch version tutorial for HoVerNet training pipeline.

### Checks


Nic-Ma:
Please add it to the root README:
https://github.com/Project-MONAI/tutorials/blob/main/README.md?plain=1#L167

Thanks.

Nic-Ma (approved):

drbeh (changes requested):
This looks good to me but before merging let me confirm with the legal the language that we used for the license.

KumoLiu:
> The notebook includes a reference to the extract_patches.py file from TIA, but the problem is that this file depends on few other files from the same TIA repo, which make it a little difficult for a user, possibly running this in Colab (assuming from the embedded URL this is the intention). I wonder whether we can extract the code needed for this operation and put it in a cell in the notebook, so that it's a bit more self contained.

Hi @JHancox, in the beginning, following the way to crop patches from the original repo was only used to reproduce their performance, if the users wanted to use their own dataset, they could have used `RandCropByPosNegLabel` or `GridPatch` these existing components in the MONAI, in which case, do we still need to add the extract the patch from the original repo?
What do you think? @drbeh
If you think it needs to be added, I will add it as soon as possible.
Thanks!

JHancox:
> > The notebook includes a reference to the extract_patches.py file from TIA, but the problem is that this file depends on few other files from the same TIA repo, which make it a little difficult for a user, possibly running this in Colab (assuming from the embedded URL this is the intention). I wonder whether we can extract the code needed for this operation and put it in a cell in the notebook, so that it's a bit more self contained.
> 
> Hi @JHancox, in the beginning, following the way to crop patches from the original repo was only used to reproduce their performance, if the users wanted to use their own dataset, they could use `RandCropByPosNegLabel` or `GridPatch` these existing components in the MONAI, in which case, do we still need to add the extract the patch from the original repo? What do you think? @drbeh If you think it needs to be added, I will add it as soon as possible. Thanks!

Thanks @KumoLiu - I certainly think it would be better to replace the reference to the TIA patch extractor with a Monai implementation within the notebook. It's not clear how to actually prepare the data at the moment.

KumoLiu:
> > > The notebook includes a reference to the extract_patches.py file from TIA, but the problem is that this file depends on few other files from the same TIA repo, which make it a little difficult for a user, possibly running this in Colab (assuming from the embedded URL this is the intention). I wonder whether we can extract the code needed for this operation and put it in a cell in the notebook, so that it's a bit more self contained.
> > 
> > 
> > Hi @JHancox, in the beginning, following the way to crop patches from the original repo was only used to reproduce their performance, if the users wanted to use their own dataset, they could use `RandCropByPosNegLabel` or `GridPatch` these existing components in the MONAI, in which case, do we still need to add the extract the patch from the original repo? What do you think? @drbeh If you think it needs to be added, I will add it as soon as possible. Thanks!
> 
> Thanks @KumoLiu - I certainly think it would be better to replace the reference to the TIA patch extractor with a Monai implementation within the notebook. It's not clear how to actually prepare the data at the moment.

Hi @JHancox, I will add it soon! Thanks!

KumoLiu:
Hi @JHancox and @drbeh, I have added extract_patches.py in https://github.com/Project-MONAI/tutorials/pull/999 and also update comments in the tutorials.  Let me know if you have any other comments, if not maybe we can merge this PR, thanks in advance!

JHancox (approved):

drbeh:
@KumoLiu in the "Check first data shape and visualize" section, the figures can be a bit misleading for the user since raw image is 270x270 but output 80x80, so it looks like that they prediction is not right when comparing side-by-side. I suggest to change the figure sizes or put a 80x80 bounding box on the raw image so users can see the prediction is related to that area. Thanks

Also it would be great to add the input image to the last image too. Thanks

drbeh:
@KumoLiu, this notebook is still using `extract_patches.py` from the original repo despite the fact that you have already implemented `prepare_patches.py` in #999. Would you please update that?

KumoLiu:
> @KumoLiu, this notebook is still using `extract_patches.py` from the original repo despite the fact that you have already implemented `prepare_patches.py` in #999. Would you please update that? Thanks

Hi @drbeh, not sure what did you mean, I have updated it in this way, "./prepare_patches.py" is pointing to the same file.
"
2. run ./prepare_patches.py to prepare patches from images. Referring to the implementation from the https://github.com/vqdang/hover_net/blob/master/extract_patches.py
"
Thanks!

drbeh:
> > @KumoLiu, this notebook is still using `extract_patches.py` from the original repo despite the fact that you have already implemented `prepare_patches.py` in #999. Would you please update that? Thanks
> 
> Hi @drbeh, not sure what did you mean, I have updated it in this way, "./prepare_patches.py" is pointing to the same file. " 2. run ./prepare_patches.py to prepare patches from images. Referring to the implementation from the https://github.com/vqdang/hover_net/blob/master/extract_patches.py " Thanks!

Oh my bad! I thought you are "referring" the users to the implementation from the https://github.com/vqdang/hover_net/blob/master/extract_patches.py

Maybe we can say "(similar to https://github.com/vqdang/hover_net/blob/master/extract_patches.py)" instead. Thanks

drbeh (approved):

PR #1055, opened 2022-11-17
ROI Inference pipeline for HoVerNet

drbeh:
Fixes https://github.com/Project-MONAI/MONAI/issues/5539

### Description

### Checks


pathology/hovernet/infer_roi.py
```diff
@@ -0,0 +1,217 @@
...
+    return log_dir
+
+
+def run(cfg):
+    # --------------------------------------------------------------------------
+    # Set Directory and Device
+    # --------------------------------------------------------------------------
+    output_dir = create_output_dir(cfg)
+    multi_gpu = True if cfg["use_gpu"] and torch.cuda.device_count() > 1 else False
+    if multi_gpu:
```
Nic-Ma, line 62: In the multi-gpu mode, I think here every GPU runs all the same data?
You should (1) partition dataset to every GPU or (2) use `DistributedSampler` in the DataLoader. Right?

Thanks.

Nic-Ma (commented):
Same as @KumoLiu 's PR, I think we should put a README doc for these python code.

Thanks.

pathology/hovernet/infer_roi.py
drbeh: Oh I know, thanks for catching that. I forgot to implement it.

pathology/hovernet/infer_roi.py
```diff
@@ -0,0 +1,217 @@
...
+    # Postprocessing transforms
+    post_transforms = Compose(
+        [
+            PromoteChildItemsd(
+                keys="pred",
+                child_keys=[HoVerNetBranch.NC.value, HoVerNetBranch.NP.value, HoVerNetBranch.HV.value],
+                delete_keys=True,
+            ),
+            Activationsd(keys=HoVerNetBranch.NC.value, softmax=True),
+            AsDiscreted(keys=HoVerNetBranch.NC.value, argmax=True),
```
KumoLiu, line 90: Seems don't need to add L89-L90, since it has been implemented here.
https://github.com/Project-MONAI/MONAI/blob/72130602340847e15f2042b2bb1b64f6f46eeb93/monai/apps/pathology/transforms/post/array.py#L660-L661
Thanks!

pathology/hovernet/infer_roi.py
drbeh: Done!

drbeh:
> Same as @KumoLiu 's PR, I think we should put a README doc for these python code.
> 
> Thanks.

I will update the same README (after #999 is merged) as they are going to be in the same folder.

KumoLiu (approved):

PR #1068, opened 2022-11-25
Py3 scripts for LUNA16 eval

Can-Zhao:
Fixes # .

### Description
Provide Py3 scripts for LUNA16 eval

### Checks


detection/run_luna16_offical_eval.sh
```diff
@@ -25,7 +25,7 @@ python ./luna16_post_combine_cross_fold_results.py \
 	-o ./result/result_luna16_all.csv
 
 mkdir -p ./result/eval_luna16_scores
-python2 ./evaluation_luna16/noduleCADEvaluationLUNA16.py \
+python ./evaluation_luna16/noduleCADEvaluationLUNA16.py \
```
mingxin-zheng, line 28: Maybe use `python3` here to distinguish it from the previous version?

detection/evaluationScript_py3_update/noduleCADEvaluationLUNA16.py
```diff
@@ -0,0 +1,515 @@
+from __future__ import print_function
+from __future__ import division
+from builtins import str
```
mingxin-zheng, line 3: Not sure if we need to import `str` and `range` explicitly

detection/run_luna16_offical_eval.sh
Nic-Ma: All the MONAI programs use python3 as default.

Thanks.

Nic-Ma (approved):

Issue #1143, opened 2023-01-06
Add tests to check formats


mingxin-zheng:
The new contributing guidelines list several requirements that need to be integrated in our GitHub actions.

**Is your feature request related to a problem? Please describe.**

Tests/actions to be added:
- [x] license/copyright check for scripts and notebooks
- [ ] standard cell block such as setup environment and setup imports
- [ ] output text limit length
- [x] DCO

mingxin-zheng:
Since most notebooks do not have the copyright headers, I think we need to make these checks optional in the beginning, revise all notebooks/source code, and finally, enforce the checks as "required".

wyli:
before finishing this ticket, I think we can enable the precommit auto formatting: https://github.com/Project-MONAI/tutorials/pull/1110. the required code change to config the tool is here https://github.com/Project-MONAI/tutorials/pull/1110/commits/6e072302688819564dd540fa06d9fdad5d9e8847

Nic-Ma:
Hi @wyli ,

Should we move the toolkit things back to this repo first[WIP]? Then auto formatting?

Thanks.

PR #1145, opened 2023-01-06
Copyright checks in notebook and script files

mingxin-zheng:
Signed-off-by: Mingxin Zheng <18563433+mingxin-zheng@users.noreply.github.com>

Adds the copyright check mentioned #1143 .

### Description

### Checks

- [x] Remove private info such as user name and private key from the notebook and script files.
- [x] Notebook runs automatically `./runner.sh -t <path to .ipynb file>`


mingxin-zheng:
The new `build / copyright ` action fails because lots of files miss copyright info, and the action plan is mentioned in #1143

runner.sh
```diff
@@ -116,6 +119,7 @@ function print_usage {
     echo "    --no-run          : don't run notebooks"
     echo "    --no-checks       : don't run code checks"
     echo "    --autofix         : autofix where possible"
+	echo "    --copyright       : check whether every source code and notebook has a copyright header"
```
wyli, line 122: there are some indentation misalignments in the script

runner.sh
```diff
@@ -227,6 +235,58 @@ if [ $doChecks = true ]; then
...
+    celltype=$4
+    ${PY_EXE} - << END
+import nbformat
+import re
+
+with open("$fname", "r") as f:
+    contents = nbformat.reads(f.read(), as_version=4)
+    cell = contents.cells[int("$cell")]
+    result = "true" if cell.cell_type == "$celltype" and re.search("$key", cell.source) else "false"
+print(result)
```
wyli, line 251: this `print` doesn't seem to print anything?

wyli (approved):
thanks, it loos good to me, feel free to revise or ignore my comments

runner.sh
mingxin-zheng: Fixed the issue to make the indentation consistent in the function, plus a few other places in my previous commit. Thanks @wyli !

runner.sh
mingxin-zheng: It prints string `"true"` or `"false"`, and the message in the stdout will be consumed by the following line:

```sh
if [[ $(verify_notebook_has_key_in_cell "$fname" "$license" 0 "markdown" ) != true ]];
```

If there are better ways to handle it, please let me know. Thanks @wyli !

wyli: I see, thanks! I think it'll be more readable to have a separate python script and a bash variable to store the returned value, like this example https://github.com/Project-MONAI/model-zoo/blob/3bffe793f13a80aa648b559d3f529bb701792ac6/ci/run_regular_tests_cpu.sh#L56-L57 but the current implementation is also ok

udiram:
@mingxin-zheng I'm seeing this problem with my PR (#1129), where it checks every file for the copyright in CI/CD, and fails. do we have to wait until all existing files are updated to include it to merge the PR?

runner.sh
mingxin-zheng: Thanks @wyli . The suggested way is implemented in #1152 .

PR #1243, opened 2023-03-07
MONAILabel: Detection and Whole Body Seg Tutorials 

tangy5:
Add Detection and Whole Body Seg Tutorials to monailabel


tangy5:
Hi @wyli @mingxin-zheng , I have been verified the new notebooks, could you have a look? If no problem, we can merge this. 
Thank you so much!

wyli (approved):
looks good to me, I can't remember why we create juypter notebooks for these documentations instead of simpler markdown files. most of them are CLI commands, I don't see how jupyter notebook makes it easier to use...cc @mingxin-zheng

tangy5:
> looks good to me, I can't remember why we create juypter notebooks for these documentations instead of simpler markdown files. most of them are CLI commands, I don't see how jupyter notebook makes it easier to use...cc @mingxin-zheng

Good point, these notebooks ideas come from the dev plan of Toolkit, then we decide to make these commands in the notebook, so that users can play with it. Another thoughts is that, in the future, these notebooks will sync with MONAI Toolkit, if markdown is better, we can quickly convert. Open discussions.

PR #1305, opened 2023-04-12
MONAI Label tutorials minor changes to the doc

tangy5:


Nic-Ma:
Hi @yiheng-wang-nv @mingxin-zheng ,

I think we don't need to introduce the different download sources and how to set them and set the authentication token in these MONAI Label tutorials, that's what the `download` API doc-string should describe clearly:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/bundle/scripts.py#L220
You can refer to other bundle related tutorials, for example:
https://github.com/Project-MONAI/tutorials/tree/main/model_zoo/adapt_bundle_to_another_dataset
If you want to introduce more, just: "for more advanced options of `download` API, please refer to: https://docs.monai.io/en/stable/bundle.html#monai.bundle.download".

What do you think?

Thanks.

yiheng-wang-nv:
Hi @Nic-Ma , agree as users may only be interested in models themselves and do not care about the source thing.

tangy5:
Hi @Nic-Ma @yiheng-wang-nv , thanks, that makes sense, removed the description of the environment.

monailabel/monailabel_pathology_nuclei_segmentation_QuPath.ipynb
```diff
@@ -106,21 +106,7 @@
     "    \n",
     "```bash\n",
     "--conf auth_token <Github Auth Token>\n",
```
Nic-Ma, line 108: Do we still need to say the `auth_token` logic?
CC @yiheng-wang-nv

tangy5: Users might face rate limit issues when using it in MONAI label. 
It's better here to remind users that they can add token as conf argument.

Nic-Ma: I mean this is a common issue of `bundle download` API, not related to MONAI Label specific logic.
We should highlight it somewhere in the `download` doc-string, and you just need to put a link here for advanced options.
@yiheng-wang-nv Could you please improve the doc-string of `download` to help users avoid this issue?

Thanks.

tangy5: That also work. Note here, MONAI Label has a way for users to provide auth_token when starting monai label server. 
This is Label specific logic, that can pass auth token to `download`. 
Let me know if we'd better remove this here.

Nic-Ma: Hi @tangy5 ,

OK, if so, to avoid too big change in MONAI Label, let's keep it here, but I would suggest to add more description to make it more clear.
How about this: `bundle.download` can support multiple sources, if downloading the bundle from Github links, user may need to provide the Github `auth_token` to avoid download limitation."
CC @yiheng-wang-nv 
What do you think?

Thanks.

yiheng-wang-nv: Hi @tangy5 , I think in MONAI label side, `auth_token` is not used for `download` API as there is no related arg.
see:
https://github.com/Project-MONAI/MONAILabel/blob/3421b4a7034514707b42efa741fa1e8903a5ba33/monailabel/utils/others/generic.py#L337

yiheng-wang-nv: `auth_token` is only used for requesting github release related information, and used in these functions:
```
get_all_bundles_list
get_bundle_versions
get_bundle_info
```
Thus the only place can be utilized for `download` is when we need to get latest version from github source.

yiheng-wang-nv: to make it more clear, let me submit a quick draft PR about it

tangy5:
Hi @yiheng-wang-nv @mingxin-zheng , I should remove all descriptions about `model-zoo`, `ngc`, `auth token`, `download source`,etc in monailabel tutorials. You could help modify other tutorials about the bundle sources. 

Thank you!

mingxin-zheng (approved):

tangy5:
are we clear about all the zoo/ngc/auth token descriptions? @yiheng-wang-nv  . Let's double-confirm within three of us, then we can merge. Thank you!

yiheng-wang-nv:
HI @tangy5 , I checked one more time and I think it's all clear.

PR #1373, opened 2023-05-09
Simplify the python bundle example

Nic-Ma:
### Description
This PR simplified the python based bundle example according to @wyli 's comments.
It removed the confusing "private" `@property`.

### Checks

### DIFF
--- a/bundle/python_bundle_workflow/scripts/inference.py
+++ b/bundle/python_bundle_workflow/scripts/inference.py
@@ -98,7 +98,10 @@ class InferenceWorkflow(BundleWorkflow):
                 value = getattr(self, f"get_{name}")()
             except AttributeError:
                 if property[BundleProperty.REQUIRED]:
-                    raise ValueError(f"unsupported property '{name}' is required in the bundle properties.")
+                    raise ValueError(
+                        f"unsupported property '{name}' is required in the bundle properties,"
+                        f"need to implement a method 'get_{name}' to provide the property."
+                    )
             self._props[name] = value
         return value
 
--- a/bundle/python_bundle_workflow/scripts/inference.py
+++ b/bundle/python_bundle_workflow/scripts/inference.py
@@ -95,23 +95,23 @@ class InferenceWorkflow(BundleWorkflow):
             self._props[name] = value
         else:
             try:
-                value = getattr(self, f"_{name}")()
+                value = getattr(self, f"get_{name}")()
             except AttributeError:
                 if property[BundleProperty.REQUIRED]:
                     raise ValueError(f"unsupported property '{name}' is required in the bundle properties.")
             self._props[name] = value
         return value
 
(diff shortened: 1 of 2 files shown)


bundle/python_bundle_workflow/scripts/inference.py
```diff
@@ -140,32 +134,27 @@ def _evaluator(self):
...
             StatsHandler(name="eval_log", output_transform=lambda x: None),
-            CheckpointLoader(load_path=self.bundle_root + "/models/model.pt", load_dict={"net": self._network_def}),
+            CheckpointLoader(load_path=self.bundle_root + "/models/model.pt", load_dict={"net": self.network_def}),
         ]
 
-    @property
     def _dataset(self):
         return Dataset(
-            data=self._dataset_data,
+            data=self.dataset_data,
```
vfdev-5, line 146: Shouldn't this be `self._dataset_data` ?

Nic-Ma: Hi @vfdev-5 ,

Here we call the `dataset_data` property, and the `_get_property()` function will call the `def _dataset_data()`.

Thanks.

bundle/python_bundle_workflow/scripts/train.py
```diff
@@ -110,27 +110,23 @@ def _get_property(self, name, property):
             self._props[name] = value
         else:
             try:
-                value = self.__getattribute__(f"_{name}")
+                value = getattr(self, f"get_{name}")()
             except AttributeError:
                 if property[BundleProperty.REQUIRED]:
                     raise ValueError(f"unsupported property '{name}' is required in the bundle properties.")
```
wyli, line 116: perhaps this error message can provide more info about how to resolve the error? like adding the property? create `get_{name}` method?

Nic-Ma: Sure, added in the latest commit.

Thanks.

bundle/python_bundle_workflow/scripts/inference.py
```diff
@@ -98,7 +98,10 @@ def _get_property(self, name, property):
                 value = getattr(self, f"get_{name}")()
             except AttributeError:
                 if property[BundleProperty.REQUIRED]:
-                    raise ValueError(f"unsupported property '{name}' is required in the bundle properties.")
+                    raise ValueError(
+                        f"unsupported property '{name}' is required in the bundle properties,"
```
vfdev-5, line 102: Nit: the word property we can replace with attribute as you removed properties ?

Nic-Ma: Hi @vfdev-5 ,

I see your point.
But we are using `property` in the context:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/bundle/properties.py#L30

Thanks.

PR #1461, opened 2023-07-17
Reduce scale factor and val loss across ranks for DDP

### LINKED ISSUE #1458: Reduce scale factor for DDP in generative model
Currently, scale factor is computed for each rank for DDP in generative model. 
It should be reduced across ranks to use the same scale factor. 
It does not cause a problem for performance for now because scale factors are all close to one in this example.
But in the future it might cause issue.

Same issue for validation loss. 

Reduce scale factors and validation loss across ranks. 

**Additional context**

Can-Zhao:
Fixes #1458 .

### Description
Reduce scale factor and val loss across ranks for DDP

### Checks


guopengf (approved):
LGTM

PR #1607, opened 2024-01-09
Add `torch.compile` benchmark tutorial

KumoLiu:
Fixes # .

### Description
Add a tutorial about how to use `torch.compile`.

### Checks


modules/torch_compile.ipynb
```diff
@@ -0,0 +1,541 @@
+{
```
ericspod, line 1: <p>I think this error indicates a real error we should solve in the environment we're using. If you can rerun in a fresh virtual environment to see if this still comes up, if so maybe there's an issue with monai-weekly and this version of Pytorch.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Project-MONAI/tutorials/pull/1607/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

ericspod, line 1: <p>I would instead show the mean of the times.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Project-MONAI/tutorials/pull/1607/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='24'/>

ericspod, line 1: <p>Could we try these graphs with logarithmic y axes instead? These have large empty spaces that are less effective in communicating information.</p><p> </p><p>Are these plots from the run in this notebook? I don't see the code for generating them. If you've run the other notebook mentioned and generated these there, I would instead plot the times captured in this notebook and mention that times versus the fast_training_tutorial.ipynb times are better.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Project-MONAI/tutorials/pull/1607/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='25'/>

ericspod, line 1: <p>"Use <code>torch.compile</code> with Engine Classes" </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Project-MONAI/tutorials/pull/1607/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='26'/>

ericspod:
You might have to look at the comments I made through ReviewNB to see where they were made, that context seems lost here!

modules/torch_compile.ipynb
KumoLiu: <p>Hi Eric, since for the compile mode the first iteration may take much more time than eager mode, so we may need more iteration to see the speedup from the mean value, so here I show the median value follow the behavior from the torch repo. The end-to-end result show the comparison from the total time.</p>

ericspod: The issue is that median in this case just happens to be showing what you want, but because the values aren't a statistical distribution it doesn't really mean much mathematically. The average of the compiled times should be much lower than uncompiled, that's what would actually be meaningful for the actual overall runtime.

KumoLiu: Ok, will update this part, thanks.

KumoLiu: I rerun in a fresh environment, and no error occurs.

KumoLiu: I have added the code for generating them and changed them to the plot captured in this notebook. And I am also using logarithmic y axes instead.

KumoLiu: Using mean instead.

ericspod (approved):
LGTM!

PR #1627, opened 2024-02-01
KiTS23 segmentation tutorial using Auto3DSeg.

myron:
This tutorial shows how to use Auto3DSeg with KiTS 2023 dataset from the MICCAI 2023 segmentation challenge.


auto3dseg/tasks/kits23/README.md
```diff
@@ -0,0 +1,225 @@
...
+class_names:
+  - { name: kidney_and_mass, index: [1,2,3] }
+  - { name: mass, index: [2,3] }
+  - { name: tumor, index: [2] }
+sigmoid: true
+
+```
+In KiTS 2023 challenge, the task is to segment 3 specific subregions: a) the first one must include all labels (kidneys, tumors, cysts) b) the second one is a union of tumors and cysts b) the third one is a tumor only region.
+Thus the "class_names" options indicates which label indices to merge together to create these 3 subregions. The "name" keys are used in Tensorboard for convenience (you can use your own names).
+Since, the 3 sub-regions are overlapping the segmentation task is a multi-label task, where each voxel can potentially be assigned to several regions. A common formulation of such task is to use the final network activation a "sigmoid", instead of the default "softmax" for mutually exclusive label classes. The config options "sigmoid: true" explicitly indicates that we are going to use the final sigmoid activation (for the multi-label segmentation). Strictly speaking it's not necessary to add this option here, the system will automatically figure it out after inspecting "class_names" having the overlapping indices.
```
KumoLiu, line 175: Does this "class_names" can be only recognized in segresnet? BTW, can you help point me out where it was identified?

myron: 1) I should work for all algos. 
2)  "indentified?"  what do you mean? in the code it's used in algo.py by each algo, and in segmenter.py for segres.
if you mean where is it defined on kits23 challenge, it's in the description https://kits-challenge.org/kits23/ under "How to win"

KumoLiu: Yes, I mean where it was used in the code.
I think it used to be a list, I didn't find where the `index` was used.
https://github.com/Project-MONAI/tutorials/blob/3093dec599f04657a7ae6c7a9c97eba5c910259a/auto3dseg/tasks/msd/Task04_Hippocampus/input.yaml#L10

KumoLiu: Found it, https://github.com/Project-MONAI/research-contributions/blob/602bbe5b95ef29d56d1a933d5bd58af3ec144f6a/auto3dseg/algorithm_templates/swinunetr/scripts/algo.py#L231
Merge this PR for now.

KumoLiu (approved):

PR #1631, opened 2024-02-04
The tutorial accompanying the pull request #7308 for MONAI core, which adds SURE-loss and Conjugate Gradient

cxlcl:
The tutorial accompanying the[ pull request #7308 ](https://github.com/Project-MONAI/MONAI/pull/7308
)for MONAI core, which adds SURE-loss and Conjugate Gradient

### Description

### Checks

- [x] Notebook runs automatically `./runner.sh -t <path to .ipynb file>`


generative/smrd/models/models.py
```diff
@@ -0,0 +1,123 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import torch.nn as nn
+from mutils import get_sigmas
+from .layers import *
+from .normalization import get_normalization
+
+
+class UNet(nn.Module):
```
KumoLiu, line 18: I think we may need to use the components from the MONAI repo.

cxlcl: Did you mean [this](https://docs.monai.io/en/1.3.0/_modules/monai/networks/nets/unet.html) one?

KumoLiu: Yes, do we have specific reason to add this pretrained model otherwise we can just ensure functionality level works?
What do you think? cc @ericspod

ericspod: It would be best wherever possible to use MONAI components. We have our own UNet class but is structured differently from this one, and we do have other blocks and layers that could be substituted for those defined here. If possible please use those in MONAI unless it causes significant headaches converting the weight names in your pretrained data.

generative/smrd/SMRD.ipynb
ericspod, line 1: I would put the code for `SMRDOptimizer` into its own file since it's so large in this notebook.

generative/smrd/models/ema.py
```diff
@@ -0,0 +1,58 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import torch.nn as nn
+
+
+class EMAHelper(object):
```
ericspod, line 15:
```suggestion
class EMAHelper:
```

generative/smrd/models/models.py
```diff
@@ -0,0 +1,123 @@
...
+        self.act = act = get_act(config)
+        self.register_buffer("sigmas", get_sigmas(config))
+        self.config = config
+
+        self.begin_conv = nn.Conv2d(config.data.channels, ngf, 3, stride=1, padding=1)
+        self.normalizer = self.norm(ngf, self.num_classes)
+
+        self.end_conv = nn.Conv2d(ngf, config.data.channels, 3, stride=1, padding=1)
+
+        self.res1 = nn.ModuleList(
```
ericspod, line 35: You could use `Sequential` instead of `ModuleList` here and elsewhere in this class, then you wouldn't need `_compute_cond_module`.

generative/smrd/models/normalization.py
```diff
@@ -0,0 +1,220 @@
...
+            out = gamma.view(-1, self.num_features, 1, 1) * x
+        return out
+
+
+class NoneNorm2d(nn.Module):
+    def __init__(self, num_features, bias=True):
+        super().__init__()
+
+    def forward(self, x):
+        return x
```
ericspod, line 159:
```suggestion
class NoneNorm2d(nn.Identity):
    def __init__(self, num_features, bias=True):
        super().__init__()
```

README.md
```diff
@@ -309,3 +309,7 @@ This tutorial shows the use cases of training and validating a 3D Latent Diffusi
 
 ##### [2D latent diffusion model](./generative/2d_ldm)
 This tutorial shows the use cases of training and validating a 2D Latent Diffusion Model.
+
+##### [SURE-Based Robust MRI Reconstruction with Diffusion Models (SMRD)](./generative/smrd)
+Example shows the use case of inference with a pre-trained score function while taking into account available measurements,
+using the SURE-Based Robust MRI Reconstruction with Diffusion Models (SMRD) method.
```
ericspod, line 315:
```suggestion
This example shows the use case of inference with a pre-trained score function while taking into account available measurements,
using the SURE-Based Robust MRI Reconstruction with Diffusion Models (SMRD) method.
```

generative/README.md
```diff
@@ -14,3 +14,7 @@ Example shows the use cases of training and validating a 3D Latent Diffusion Mod
 
 ## [Brats 2D latent diffusion model](./2d_ldm/README.md)
 Example shows the use cases of training and validating a 2D Latent Diffusion Model on axial slices from Brats 2016&2017 data.
+
+## [SURE-Based Robust MRI Reconstruction with Diffusion Models (SMRD)](./smrd/README.md)
+Example shows the use case of inference with a pre-trained score function while taking into account available measurements,
+using the SURE-Based Robust MRI Reconstruction with Diffusion Models (SMRD) method.
```
ericspod, line 20:
```suggestion
This example shows the use case of inference with a pre-trained score function while taking into account available measurements,
using the SURE-Based Robust MRI Reconstruction with Diffusion Models (SMRD) method.
```

ericspod:
Thanks for the contribution @cxlcl.

There's a lot of code in this tutorial to load your pretrained model. I went through it a bit with some minor suggestions but not in great detail. If it's possible to save your pretrained model as a Torchscript object we can do away with the model code in this tutorial and point people to the repo for the paper if they want to see what the model is. If not then, wherever possible without breaking your saved weights too much, please replace your layer/block definitions with those from Core to reduce overall code volume. 

I would still move the code for the optimizer to its own file so that the notebook isn't huge, unless you can reduce its size given the comments I made. The notebook would benefit with more description of what's going on with this technique. 

There's quite a few things here but it looks like a good addition overall. The CI fails aren't major issues we can fix later. 

@marksgraham you may want to look over this as well.

Thanks!

marksgraham:
> Thanks for the contribution @cxlcl.
> 
> There's a lot of code in this tutorial to load your pretrained model. I went through it a bit with some minor suggestions but not in great detail. If it's possible to save your pretrained model as a Torchscript object we can do away with the model code in this tutorial and point people to the repo for the paper if they want to see what the model is. If not then, wherever possible without breaking your saved weights too much, please replace your layer/block definitions with those from Core to reduce overall code volume.
> 
> I would still move the code for the optimizer to its own file so that the notebook isn't huge, unless you can reduce its size given the comments I made. The notebook would benefit with more description of what's going on with this technique.
> 
> There's quite a few things here but it looks like a good addition overall. The CI fails aren't major issues we can fix later.
> 
> @marksgraham you may want to look over this as well.
> 
> Thanks!

Looks good to me, I think this stuff is complementary to the diffusion work we're doing. As already mentioned by others, would have been great to use the monai networks already defined for the pretrained network (would cut down on the code added and make it easier for others to integrate this into their own work). I wonder if it would be hard to re-train the network using a MONAI implementation and then integrate that? Might be easier than trying to convert weights over

cxlcl:
Thanks a lot for all the suggestions. Based on the suggestions, I've made several changes:

1. Removed the dependency on the UNet structure definition on code files. Following @ericspod 's suggestion, the model definition are saved in TorchScript. So any user can load it directly without relying on additional files. This reduces the code size for the tutorial. The pointer to the definition for the UNet is added in case the user is interested.  After a closer look at the layers in the pre-trained model, I found it is not feasible to replace the layers there with the ones in MONAI, since the ResBlock and RefineBlock within their UNet is different from the ones in MONAI. Rewrite those layers end up with writing costumed ResBlock and RefineBlock.

2. Revised the notebook. The notebook is revised such that each of the Conjugate Gradient and SURE loss has a toy example, illustrating what it does. Then the two modules are combined in the `SMRDoptimizer`, in a separate file. More descriptions are added to illustrate the basic ideas.

3. Minor changes to the code, followed the suggestions from the revision

Please let me know how it looks and thanks again for providing the feedbacks!

generative/smrd/models/ema.py
```diff
@@ -0,0 +1,58 @@
...
+        for name, param in module.named_parameters():
+            if param.requires_grad:
+                self.shadow[name].data = (1.0 - self.mu) * param.data + self.mu * self.shadow[name].data
+
+    def ema(self, module):
+        if isinstance(module, nn.DataParallel):
+            module = module.module
+        for name, param in module.named_parameters():
+            if param.requires_grad:
+                param.data.copy_(self.shadow[name].data)
```
ericspod, line 39: A suggested shortening:

```suggestion
class EMAHelper:
    def __init__(self, mu=0.999):
        self.mu = mu
        self.shadow = {}
        
    def _get_parameters(self,module):
        if isinstance(module, nn.DataParallel):
            module = module.module
        return module.named_parameters()

    def register(self, module):
        for name, param in self._get_parameters(module):
            if param.requires_grad:
                self.shadow[name] = param.data.clone()

    def update(self, module):
        for name, param in self._get_parameters(module):
            if param.requires_grad:
                self.shadow[name].data = (1.0 - self.mu) * param.data + self.mu * self.shadow[name].data

    def ema(self, module):
        for name, param in self._get_parameters(module):
            if param.requires_grad:
                param.data.copy_(self.shadow[name].data)
```

generative/smrd/models/ema.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def ema_copy(self, module):
+        if isinstance(module, nn.DataParallel):
+            inner_module = module.module
+            module_copy = type(inner_module)(inner_module.config).to(inner_module.config.device)
```
ericspod, line 44: Can we not do `inner_module.clone()`? This line will work if the type of the inner module is something constructable like this, but the Pytorch pattern is to use the clone method then the `load_state_dict()` method call isn't needed.

ericspod:
This looks much better and much smaller. I had a few comments in the notebook and code but overall it's looking great. The CI fails can be fixed as well, it's too bad this can't be tested with papermill at the moment but we can live with it.

cxlcl:
<p>Added in the latest commit. Forgot to include it in the previous commit... Thanks for the comments for the code style. I will address it later this week.</p>

---

 <a href='https://app.reviewnb.com/Project-MONAI/tutorials/pull/1631/discussion/#comment-1965998913'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

ericspod:
Hi @cxlcl sorry for losing track on this PR. Could you please fix the DCO issue and see if there's anything in the unresolved comments that need to be addressed still? Thanks!

ericspod:
Hi @cxlcl if you're still wanting to finish this tutorial we can definitely review once the issues are sorted. Things have changed a bit so the conflicts and other things will have to be sorted before review and merging. Thanks!

cxlcl:
Hi Eric, Sorry for the late response. Yes, I would like to resolve the DCO issue. I've tried to resolve it by first run the local check script and upload. But the issue is still there in the CI . Is there any other way to check it locally or quickly without waiting for the results from the online CI process?

PR #1654, opened 2024-02-26
Update compile benchmark tutorial

KumoLiu:
Fixes # .

### Description
Revert logarithmic y axes plt in compile result show.

### Checks


KumoLiu:
Hi @ericspod, following the suggestions you mentioned in this [comment](https://github.com/Project-MONAI/tutorials/pull/1607#discussion_r1457261611), I revised the graphs and substituted with logarithmic y-axes. However, after an offline discussion with @Nic-Ma, we're of the opinion that the use of logarithmic y-axes might introduce ambiguity in illustrating the speed enhancement of the actual compile relative to eager training. This could potentially lead users to overestimate the magnitude of the speed increase. What do you think?

ericspod:
Hi @KumoLiu, I see the concern with logarithmic y axis. My feeling is still that the graphs as they are don't convey the benchmark information well but I'm trying to figure out what would. I wonder if presenting the compiled values as a ratio of the baseline uncompiled times would help, especially in the right hand graph if you omit the first epoch or two which have those spikes. If you're both okay with this tutorial as it is we can come back to this some other time, the demonstration value of the notebook is good to have in the tutorials.

KumoLiu:
> I wonder if presenting the compiled values as a ratio of the baseline uncompiled times would help, especially in the right hand graph if you omit the first epoch or two which have those spikes.

Hi @ericspod, yes. If you're training a rather small model for a limited number of epochs, the compile feature may not significantly boost the speed. On the flip side, when it comes to larger scale training, the speedup gains from the compile feature tend to become more evident. I believe that detailing the execution times for the initial few epochs within the notebook would offer users insight into this behavioral pattern. What do you think?

KumoLiu:
Alternatively, we can keep the graph of the logarithmic y axis as well.

Nic-Ma (approved):

KumoLiu:
Merged for now, and we can come back to this some other time when we find a better way.

PR #1656, opened 2024-02-28
Update `python_bundle_workflow`

KumoLiu:
Fixes # .

### Description

- add description for `self._set_prop` in python `python_bundle_workflow`.
- remove generate data outside of the Workflow class.

### Checks


bundle/python_bundle_workflow/scripts/inference.py
```diff
@@ -63,7 +63,6 @@ def __init__(self, dataset_dir: str = "."):
             nib.save(n, os.path.join(dataset_dir, f"seg{i:d}.nii.gz"))
 
         self._props = {}
-        self._set_props = {}
         self.dataset_dir = dataset_dir
 
     def initialize(self):
```
ericspod, line 68: This should also set `self._props`?

ericspod (approved):
I guess this is fine to change unless there's a use for `_set_props` I don't see. @Nic-Ma any thoughts? Thanks!

bundle/python_bundle_workflow/scripts/inference.py
KumoLiu: I have removed this line, as it appears to overwrite the probs that have already been set in the init.

Nic-Ma (changes requested):
Hi @KumoLiu @ericspod ,

I remember the idea of `_set_props` is to save the instantiated python objects, so if some python object is not changed and you can `initialize()` again, it can reuse it.
What do you think?

Thanks.

ericspod:
> Hi @KumoLiu @ericspod ,
> 
> I remember the idea of `_set_props` is to save the instantiated python objects, so if some python object is not changed and you can `initialize()` again for some changed objects, it can reuse it. What do you think?
> 
> Thanks.

Hi @Nic-Ma is this being used somewhere or will be? If it is we should write a bit more commentary since the semantics wasn't so clear. If not and won't ever be used I would take it out. Thanks!

Nic-Ma:
> > Hi @KumoLiu @ericspod ,
> > I remember the idea of `_set_props` is to save the instantiated python objects, so if some python object is not changed and you can `initialize()` again for some changed objects, it can reuse it. What do you think?
> > Thanks.
> 
> Hi @Nic-Ma is this being used somewhere or will be? If it is we should write a bit more commentary since the semantics wasn't so clear. If not and won't ever be used I would take it out. Thanks!

Hi @ericspod ,

There is a typical use-case: if we want to set the bundle idle on GPU and run inference for every input image, we can only override the `data` part and re-initialize the workflow, it can help avoid recreating the model and loading weights file to GPU.
@KumoLiu Can we add some comments in the example to explain?
What do you think?

Thanks.

KumoLiu:
Hi @ericspod,
Following an offline discussion with Nic, I've updated the tutorial. The `self._set_props` is utilized to store user-reset initialized objects, avoiding unnecessary re-initializations. 

To aid understanding and also added as a note, I've provided a brief example.
```
wf = TrainWorkflow()
wf.initialize()
wf.run()
wf.loss = DiceFocalLoss(sigmoid=True)
# The workflow requires re-initialization, but the user-reset loss should remain as is.
wf.initialize()
```

bundle/python_bundle_workflow/scripts/inference.py
KumoLiu: I have added back and updated to set `self._props`.

bundle/python_bundle_workflow/scripts/inference.py
```diff
@@ -112,7 +108,7 @@ def get_device(self):
         return torch.device("cuda" if torch.cuda.is_available() else "cpu")
 
     def get_dataset_dir(self):
-        return "."
+        return "./infer"
```
ericspod, line 111:
```suggestion
        return self.dataset_dir
```

bundle/python_bundle_workflow/scripts/train.py
```diff
@@ -48,32 +49,36 @@
...
-            n = nib.Nifti1Image(im, np.eye(4))
-            nib.save(n, os.path.join(dataset_dir, f"img{i:d}.nii.gz"))
-            n = nib.Nifti1Image(seg, np.eye(4))
-            nib.save(n, os.path.join(dataset_dir, f"seg{i:d}.nii.gz"))
+        prepare_data(dataset_dir=dataset_dir)
 
         # define buckets to store the generated properties and set properties
         self._props = {}
         self._set_props = {}
-        self.dataset_dir = dataset_dir
```
ericspod, line 76: Shouldn't this be kept and `get_data_dir()` return it?

ericspod (approved):
I have made a comment about the swapped order of the if-conditions in both classes, but the other changes look good to me so I can approve now and leave it up to you to swap the order of conditions or not. Thanks!

bundle/python_bundle_workflow/scripts/train.py
```diff
@@ -103,11 +110,11 @@ def _get_property(self, name, property):
...
         """
         value = None
-        if name in self._props:
-            value = self._props[name]
-        elif name in self._set_props:
+        if name in self._set_props:
             value = self._set_props[name]
             self._props[name] = value
+        elif name in self._props:
+            value = self._props[name]
```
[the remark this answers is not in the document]
KumoLiu: Yes, it appears set_props should have higher priority than _props.
cc @Nic-Ma double confirm.

PR #1673, opened 2024-03-20
Update "pathology/tumor_detection/ignite/profiling_camelyon_pipeline.ipynb"

KumoLiu:
Fixes # .

### Description
Update the usage of nvtx in "tumor_detection_camelyon"

### Checks


drbeh (approved):
Thanks @KumoLiu! I have tested it on my machine and all works fine.

drbeh:
@KumoLiu I see an error in the CI but I don't see it when running locally. Can it be due the installed nsight version?

KumoLiu:
It’s wired, I don’t see any reason why the file is not exist. Also, what could be the reasons for it not executing immediately.

ericspod:
I would think papermill would run all command invocations otherwise many other notebooks would fail. How large is the generated csv? If the container runs out of space and tries to write the file it may silently do nothing, but this file would have to be really big. We could have a check right after `nsys stats` to check that the file was created: `assert os.path.isfile("profile_report_nvtx_pushpop_sum.csv")`. For testing in the CICD environment we could temporarily add `!ls -lh` right before that to see what the file structure looks like.

KumoLiu:
works well in the 24.02 PyTorch image, close for now.

PR #1707, opened 2024-05-07
remove plastimatch package in tutorial

tangy5:


monailabel/monailabel_radiology_spleen_segmentation_OHIF.ipynb
```diff
@@ -124,22 +124,9 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "### 3.2 Convert to Dicom Format\n",
+    "### 3.2 Prepare Dicom Data\n",
```
ericspod, line 127: Could we still provide this step to users with another tool instead of Plastimatch which doesn't require apt? For example, https://github.com/ENHANCE-PET/nifti2dicom

KumoLiu: Hi @tangy5, I tried with pyplastimatch which is a python wrapper for plastimatch https://github.com/ImagingDataCommons/pyplastimatch, it works well. 

Perhaps let's using this package as an example?
```
import pyplastimatch as pypla
from pyplastimatch.utils.install import install_precompiled_binaries

install_precompiled_binaries()

convert_args_ct = {
    "input" : "image.nii.gz",
    "patient-id" : 'patient1',
     "output-dicom": "test"
                  }
pypla.convert(verbose = True, **convert_args_ct)
```

Thanks.

tangy5: Thanks, this is great.

KumoLiu (approved):

PR #1795, opened 2024-08-21
Update luna detection

KumoLiu:
Try to use `ApplyTransformToPointsd` to update luna detection pipeline

### Checks


detection/generate_transforms.py
```diff
@@ -191,6 +201,156 @@ def generate_detection_train_transform(
...
+
+    Return:
+        training transform for detection
+    """
+    if amp:
+        compute_dtype = torch.float16
+    else:
+        compute_dtype = torch.float32
+
+    train_transforms = Compose(
```
KumoLiu, line 240: Hi @ericspod @atbenmurray @Nic-Ma and @mingxin-zheng, could you please help take a look whether this new transform pipeline make sense to you. I test it in this test [notebook](https://github.com/Project-MONAI/tutorials/blob/002153ab02e408c64056221e6cc5ce343134f1c6/detection/test-identical.ipynb), it achieve the same result as before.

detection/generate_transforms.py
```diff
@@ -191,6 +201,156 @@ def generate_detection_train_transform(
...
+                keys=[point_key], refer_key=image_key, mode="world_to_image", affine_lps_to_ras=affine_lps_to_ras
+            ),
+            ConvertPointsToBoxd(keys=[point_key]),
+            ClipBoxToImaged(
+                box_keys=box_key,
+                label_keys=[label_key],
+                box_ref_image_keys=image_key,
+                remove_empty=True,
+            ),
+            BoxToMaskd(
```
KumoLiu, line 311: I leave this `BoxToMaskd`, `RandRotated` and `MaskToBoxd` here to achieve the same result as before.
It can also be implemented by using `CoordinateTransformd` but the issue here is that `BoxToMaskd` not only convert the bbox to the mask, it will also try to consider if the label is more like a ellipse, it will try to generate a fake ellipse mask.

mingxin-zheng: According to the discussion yesterday, v1 seems to provide identical results, but v2 is more native and accumulate less errors in the processing pipeline. Am I right or wrong here @KumoLiu ?

KumoLiu: Yes, v1 can provide identical results but v2 will be more easy to use. But I would suggest using v1 in this version since I didn't using v2 to run the end-to-end pipeline so I can't ensure whether it will influence the performance.

mingxin-zheng: Okay I see. That makes sense to me

KumoLiu:
I didn't modify the `generate_detection_val_transform` and `generate_detection_inference_transform` because it would offer no significant benefit. Making these changes would require converting boxes to points, applying the transform, and then converting the points back to boxes.

detection/generate_transforms.py
```diff
@@ -39,12 +46,18 @@
...
+from monai.data.box_utils import clip_boxes_to_image
+from monai.apps.detection.transforms.box_ops import convert_box_to_mask
+from monai.transforms.utils_pytorch_numpy_unification import stack, concatenate, min, max
 
 
 def generate_detection_train_transform(
     image_key,
     box_key,
     label_key,
+    point_key,
```
mingxin-zheng, line 55: Hi @KumoLiu , would it make sense if we give a default value (`points`) to the `point_key`? Some users may not be interested to understand  what are the names of the points.

detection/generate_transforms.py
```diff
@@ -307,3 +318,65 @@ def generate_detection_inference_transform(
...
+    """
+
+    def __init__(self, keys: KeysCollection, image_key: str, spatial_size, whole_box, mask_image_key="mask_image"):
+        """
+        Args:
+            keys: keys of the corresponding items to be transformed.
+            image_key: key for the image data in the dictionary.
+            box_key: key for the box data in the dictionary.
+            mask_key: key to store the generated box mask.
+        """
```
mingxin-zheng, line 330:
```suggestion
    def __init__(self, keys: KeysCollection, image_key: str, spatial_size: Tuple[int, int, int], whole_box: bool, mask_image_key: str = "mask_image") -> None:
        """
        Args:
            keys: keys of the corresponding items to be transformed.
            image_key: key for the image data in the dictionary.
            spatial_size: size of the spatial dimensions of the mask.
            whole_box: whether to use the whole box for generating the mask.
            mask_image_key: key to store the generated box mask.
        """
```

detection/generate_transforms.py
```diff
@@ -307,3 +318,65 @@ def generate_detection_inference_transform(
...
+        )
+        mask_img = np.amax(mask_img, axis=0, keepdims=True)[0:1, ...]
+        return mask_img
+
+    def __call__(self, data):
+        d = dict(data)
+        for key in self.key_iterator(d):
+            image = d[self.image_key]
+            boxes = d[key]
+            data[self.mask_image_key] = self.generate_mask_img(boxes, image.shape[1:], self.whole_box)
```
mingxin-zheng, line 376:
```suggestion
            data[self.mask_image_key] = self.generate_mask_img(boxes, image.shape[1:], whole_box=self.whole_box)
```

detection/luna16_training.py
```diff
@@ -233,7 +236,7 @@ def main():
     )
     after_scheduler = torch.optim.lr_scheduler.StepLR(optimizer, step_size=150, gamma=0.1)
     scheduler_warmup = GradualWarmupScheduler(optimizer, multiplier=1, total_epoch=10, after_scheduler=after_scheduler)
-    scaler = torch.cuda.amp.GradScaler() if amp else None
+    scaler = torch.amp.GradScaler('cuda') if amp else None
```
KumoLiu, line 239: Update to remove this warning.
```
WARNING:py.warnings:`torch.cuda.amp.GradScaler(args...)` is deprecated. Please use `torch.amp.GradScaler('cuda', args...)` instead.
WARNING:py.warnings:`torch.cuda.amp.autocast(args...)` is deprecated. Please use `torch.amp.autocast('cuda', args...)` instead.
```

detection/luna16_prepare_images.py
```diff
@@ -47,8 +47,10 @@ def main():
 
     monai.config.print_config()
 
-    env_dict = json.load(open(args.environment_file, "r"))
-    config_dict = json.load(open(args.config_file, "r"))
+    with open(args.environment_file, "r") as env_file:
```
KumoLiu, line 50: Update to remove this warning.
```
WARNING:py.warnings:unclosed file <_io.TextIOWrapper name='./config/environment_luna16_fold0.json' mode='r' encoding='UTF-8'>
WARNING:py.warnings:unclosed file <_io.TextIOWrapper name='./config/config_train_luna16_16g.json' mode='r' encoding='UTF-8'>
```

detection/generate_transforms.py
KumoLiu: Yes! Add a default value for point_key.

mingxin-zheng:
Looks good to me overall. Hi @Can-Zhao , we introduced some early features about point operations. Would you mind taking a look at this PR? Thanks!

detection/generate_transforms.py
Can-Zhao: LGTM. Thanks.

Can-Zhao (approved):
Looks great. Thanks!

PR #1898, opened 2024-12-15
Call Vista3D NIM on NVIDIA AI enterprise / GPU cloud

aylward:
Fixes # .

### Description
Illustrates functions for uploading images to file.io and using their URLs to launch Vista3D NIM on NVIDIA AI enterprise / GPU cloud servers.

Enables researchers with limited GPU resources to take advantage of NVIDIA resources while evaluating MONAI Vista3D.

### Checks

- [X] Clean up long text outputs from code cells in the notebook.
    - **Notebook requires a private API key.**  Anyone can register to get one, but notebook will not automatically run without one.

### DIFF
--- a/nvidia_nims/vista_3d_remote_nim.ipynb
+++ b/nvidia_nims/vista_3d_remote_nim.ipynb
@@ -1,0 +1,464 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "7b50dffd",
+   "metadata": {},
+   "source": [
+    "```\n",
+    "Copyright (c) MONAI Consortium\n",
+    "Licensed under the Apache License, Version 2.0 (the \"License\");\n",
+    "you may not use this file except in compliance with the License.\n",
+    "You may obtain a copy of the License at\n",
+    "   http://www.apache.org/licenses/LICENSE-2.0\n",
+    "Unless required by applicable law or agreed to in writing, software\n",
+    "distributed under the License is distributed on an \"AS IS\" BASIS,\n",
+    "WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.\n",
+    "See the License for the specific language governing permissions and\n",
+    "limitations under the License.\n",
+    "```\n",
+    "\n",
+    "# Calling Vista-3D NIM on NVAIE\n",
+    "\n",
+    "In this tutorial, we will cover the following:\n",
+    "\n",
+    "- **Convert image to Nifty format:** The Vista-3D NIM requires files to be in compressed nifty (.nii.gz) format.  We use ITK to convert DICOM and other file formats to nifty.\n",
+    "\n",
+    "- **Create a single-use URL to the image:** We use the website file.io to provide a single-use URL for our image.  The Vista-3D NIM requires a URL to the image to be processed.  File.io provides an API for receiving a file and returning a one-time use URL to that file.\n",
+    "\n",
+    "- **Communicate with the Vista-3D NIM on NVAIE:** The Vista-3D NIM can be downloaded and run locally, but herein we are running it on the NVIDIA AI Enterprise (NVAIE) / NVIDIA GPU Cloud (NGC) servers.  You can get free NVAIE/NGC credits and an API Key for remotely accessing those servers by registering at https://build.nvidia.com/nvidia/vista-3d.  Also, you can apply for unlimited access as a researcher by joining the (free) NVIDIA Developer Program: https://developer.nvidia.com/join-nvidia-developer-program\n",
+    "\n",
+    "- **Visualize and/or save the results:** We provide pyvista for visualization and use ITK to save the results in any of a variety of image formats.\n",
+    "\n",
+    "Learn more about the Vista-3D foundation model for 3D CT segmentation:<br>\n",
+    "https://arxiv.org/abs/2304.02643\n",
+    "\n",
+    "Learn more about NIMs:<br>\n",
+    "https://www.nvidia.com/en-us/ai/"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "b1fc03f5",
+   "metadata": {},
+   "source": [
+    "\n",
+    "## Obtaining an NVIDIA AI Enterprise / NVIDIA GPU Cloud (NVAIE/NGC) API Key\n",
+    "\n",
+    "This demonstration runs the Vista-3D NIM (AI container) on the NVIDIA AI Enterprise servers, so an API key is required.  The good news is that any individual can get an API key and 1,000 to 5,000 free credits (processing one image requires one credit) when they first sign-up at:\n",
+    "\n",
+    "- https://build.nvidia.com/nvidia/vista-3d\n",
+    "\n",
+    "Also, you can apply for unlimited local access as a researcher by joining the (free) NVIDIA Developer Program:\n",
+    "\n",
+    "- https://developer.nvidia.com/join-nvidia-developer-program\n",
+    "\n",
+    "This API Key is specific to your account at NVIDIA.\n",
+    "\n",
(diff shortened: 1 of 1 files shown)


nvidia_nims/vista_3d_remote_nim.ipynb
```diff
@@ -0,0 +1,510 @@
+{
```
KumoLiu, line 1: <pre class="ql-syntax" spellcheck="false">Line #2.    !python -c "import pyvista" || ipywidgets 'pyvista[all,trame]'

</pre><p> </p><p>Looks like a typo? Should be <code>!python -c "import pyvista" || pip install -q 'pyvista[all,trame]'</code></p><p> </p><p>BTW, pyvista 0.44.2 does not provide the extra 'trame', "all" should be enough here?</p><p><code>WARNING: pyvista 0.44.2 does not provide the extra 'trame'</code></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Project-MONAI/tutorials/pull/1898/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

KumoLiu, line 1: <p>Do you think adding two screenshots here would improve the overall clarity and visual appeal?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Project-MONAI/tutorials/pull/1898/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='19'/>

KumoLiu (approved):
Thanks for the PR! Looks good to me.
Put two minor comments.

nvidia_nims/vista_3d_remote_nim.ipynb
aylward: Fixed.

aylward: Added.

KumoLiu: You may need using `Figure`, the screenshots not shown as expected here: https://github.com/Project-MONAI/tutorials/blob/95eff84dfbadcfba066c3b09e88176d905e1aff7/nvidia_nims/vista_3d_remote_nim.ipynb

nvidia_nims/vista_3d_remote_nim.ipynb
```diff
@@ -0,0 +1,523 @@
...
```
ericspod, line 1: <pre class="ql-syntax" spellcheck="false">Line #5.        invoke_url = "https://health.api.nvidia.com/v1/medicalimaging/nvidia/vista-3d"
</pre><p>I would turn this into an argument for the function, and where you call this function pass in this URL explaining what it is, and maybe suggest other models that would work here too to do something different. I think that helps illustrate what the end point is in the notebook by explicitly using it in a cell rather than being hard coded in a function. At the top of the notebook I would also provide very brief descriptions of what VISTA 3D and NIM are, but very brief next to the URLs you provide for each. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Project-MONAI/tutorials/pull/1898/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

ericspod, line 1: <pre class="ql-syntax" spellcheck="false">Line #9.        isotropy = (
</pre><p>Here I'd recommend saving the spacing to a variable: <code>input_spacing = nput_image.GetSpacing()</code> and then use that variable everywhere you're calling <code>input_image.GetSpacing()</code>. This reduces the code a bit and helps with readability.  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Project-MONAI/tutorials/pull/1898/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

ericspod, line 1: <pre class="ql-syntax" spellcheck="false">Line #55.                    return itk.imread(filepath, pixel_type=itk.SS)
</pre><p>if <code>imread</code> can read from a data stream you can use <code>z.open</code> to open streams to files in the zip file without extracting it, for large zips this can be faster.  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Project-MONAI/tutorials/pull/1898/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

ericspod, line 1: <pre class="ql-syntax" spellcheck="false">Line #1.    ngc_api_key = os.environ.get("NGC_API_KEY")
</pre><p>It may be worth noting here that people can just paste their API keys here if they're in an environment that setting an variable isn't feasible.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Project-MONAI/tutorials/pull/1898/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

ericspod:
I had some super minor comments on the notebook but otherwise looks good to me. We can't see the PyVista widgets in the notebook when rendered in Github, I wonder if adding the screenshots in their place as dummies would work. What you could do is run a rendering cell with the code commented and instead displaying an image, then restore the content of the cell with the code before saving the notebook.

nvidia_nims/vista_3d_remote_nim.ipynb
aylward: Quite a bit of this function is specific to the VISTA 3D NIM and this web interface/URL (e.g., the validation logic and the URL request parameters/arguments).   So, I don't think the URL should be a parameter to this function.

However, I agree that a brief description of VISTA 3D and NIMs are needed, and I will add them.

aylward: Streaming from zips is already done by ITK (i.e., imread) depending on the file format (e.g., you can stream MetaIO files).   It also has a very fast way of parsing directories of DICOM objects to load 3D volumes (by default loading the first series in the directory, if a series ID isn't specified).

I am not certain if ITK provides a way of specifying a collection of DICOM via gzip - I know we have written programs on top of ITK that do that, but interesting idea to add that capability to ITK's imread directly.  As you note, it would be easy.    It seems outside the scope of this notebook, or am I missing something?

aylward: Ok - I am willing to add that, as long as we also note that it is bad practice :).   Too often people forget to delete the keys before checking in their code, thereby inadvertently advertising their keys.  We shouldn't encourage that practice, imo.

ericspod: <p>Sounds good!</p>

ericspod: <p>Yeah it's bad practice for sure but we can note that very clearly, I'm thinking of cases like Colab where environment variables aren't so easily set.</p>

aylward: <p>Thought about it more, and used your idea of making the URL an argument.   Thanks for the suggestion!</p>

aylward: <p>Good point about Colab.   Added comment.</p>

aylward:
File.io was bought by limewire last week, and they have shut down the file.io service.  I have an alternative solution using Google Cloud, but it isn't an easy lift - requiring someone to sign up for a free gcloud account, if they don't already have one.    I'm working with the Vista3D NIM folks to find an alternative solution.  Stay tuned...

PR #1907, opened 2024-12-20
Added TRT config for inference

borisfom:
### Description
Added TRT config for MAISI and the extension of the inference script to handle extra config file.

Note: autoencoder.decoder currently cannot  be exported to TRT (crashes during engine generation).
It does not seem to take a big part of the whole run anyway.


generation/maisi/configs/config_trt.json
```diff
@@ -0,0 +1,22 @@
...
+                -1
+            ],
+            [
+	    ]
+        ]
+    },
+    "device": "cuda",
+    "controlnet": "$trt_compile(@controlnet_def.to(@device), @trained_controlnet_path, @c_trt_args)",
+    "diffusion_unet": "$trt_compile(@diffusion_unet_def.to(@device), @trained_diffusion_path)",
+    "mask_generation_diffusion": "$trt_compile(@mask_generation_diffusion_def.to(@device), @trained_mask_generation_diffusion_path)"
```
yiheng-wang-nv, line 21: Hi @borisfom , thanks for the changes. 
Can I confirm if `mask_generation_diffusion` supports TRT? Should we also modify https://github.com/Project-MONAI/model-zoo/blob/dev/models/maisi_ct_generative/configs/inference_trt.json ?

borisfom: Well mask_generation_diffusion was never called - it probably is TRT compliant, but never tested. 
No need to modify current model-zoo config until we have a config that engages this component.

borisfom:
@yiheng-wang-nv : can we gat back to it ?

yiheng-wang-nv (approved):
Thanks @borisfom . Could you add one line in README to show that how to run inference with TRT? Then we can merge

borisfom:
@yiheng-wang-nv : I added README bits, can we get this merged ?

yiheng-wang-nv (approved):

borisfom:
@yiheng-wang-nv : Can you merge it? I don't have write access.

Issue #1865, opened 2024-10-21
Add a inference acceleration tutorial


yiheng-wang-nv:
Like what we did for training acceleration:

https://github.com/Project-MONAI/tutorials/blob/main/acceleration/fast_training_tutorial.ipynb

ericspod:
We can consider this for 1.5. This would use ONNX, TRT, something else you're thinking of?

yiheng-wang-nv:
Hi @ericspod , thanks for the suggestions!

Now I'm starting this work, the following features are considered to use:
1. GDS for data loading
2. do GPU transform
3. use TRT for inference

yiheng-wang-nv:
Except for the above features, @Nic-Ma also suggests to add quantization support, we can conduct adding this feature in the next version.

PR #1948, opened 2025-02-28
add fast inference tutorial

yiheng-wang-nv:
Fixes # .

### Description

### Checks

### DIFF
--- a/acceleration/fast_inference_tutorial/fast_inference_tutorial.ipynb
+++ b/acceleration/fast_inference_tutorial/fast_inference_tutorial.ipynb
@@ -1,0 +1,336 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Copyright (c) MONAI Consortium  \n",
+    "Licensed under the Apache License, Version 2.0 (the \"License\");  \n",
+    "you may not use this file except in compliance with the License.  \n",
+    "You may obtain a copy of the License at  \n",
+    "&nbsp;&nbsp;&nbsp;&nbsp;http://www.apache.org/licenses/LICENSE-2.0  \n",
+    "Unless required by applicable law or agreed to in writing, software  \n",
+    "distributed under the License is distributed on an \"AS IS\" BASIS,  \n",
+    "WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  \n",
+    "See the License for the specific language governing permissions and  \n",
+    "limitations under the License.\n",
+    "\n",
+    "# Fast Inference with MONAI features"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This tutorial demonstrates the performance comparison between a standard PyTorch training program and a MONAI-optimized inference program. The key features include:\n",
+    "\n",
+    "1. **Direct Data Loading**: Load data directly from disk to GPU memory, minimizing data transfer time and improving efficiency.\n",
+    "2. **GPU-based Preprocessing**: Execute preprocessing transforms directly on the GPU, leveraging its computational power for faster data preparation.\n",
+    "3. **TensorRT Inference**: Utilize TensorRT for running inference, which optimizes the model for high-performance execution on NVIDIA GPUs.\n",
+    "\n",
+    "This tutorial is modified from the `TensorRT_inference_acceleration` tutorial."
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Setup environment"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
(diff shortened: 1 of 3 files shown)


ericspod:
This addresses https://github.com/Project-MONAI/tutorials/issues/1865 I assume.

yiheng-wang-nv:
HI @ericspod @Nic-Ma @KumoLiu , the tutorial is almost ready, could you help to review the layout first?
I will add analyze and visualization results later

yiheng-wang-nv:
> HI @ericspod @Nic-Ma @KumoLiu , the tutorial is almost ready, could you help to review the layout first? I will add analyze and visualization results later

tutorial is ready.

acceleration/fast_inference_tutorial/fast_inference_tutorial.ipynb
```diff
@@ -0,0 +1,635 @@
+{
```
KumoLiu, line 1: <p>Do you think we can also include the <code>.nii.gz</code> benchmark result in the notebook since the original data is <code>nii.gz</code> format.  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Project-MONAI/tutorials/pull/1948/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

yiheng-wang-nv: Hi @KumoLiu , thanks for the suggestion. `.nii.gz` files have to be decompressed in CPU, thus using GDS may not have acceleration. I added a section to introduces the limitations on each feature, could you help to review the updates? Thanks!

acceleration/fast_inference_tutorial/fast_inference_tutorial.ipynb
```diff
@@ -0,0 +1,646 @@
...
+    "    trt_flag = True\n",
+    "if \"gpu_transforms\" in benchmark_type:\n",
+    "    gpu_transforms_flag = True\n",
+    "if \"gds\" in benchmark_type:\n",
+    "    gpu_loading_flag = True\n",
+    "\n",
+    "infer_transforms = get_transforms(device, gpu_loading_flag, gpu_transforms_flag)\n",
+    "model = get_model(device, weights_path, trt_model_path, trt_flag)\n",
+    "\n",
+    "# total_time_dict = run_inference(train_files, infer_transforms, model, device, benchmark_type)\n",
```
Nic-Ma, line 426: Why comment these lines?

Thanks.

Nic-Ma (approved):
Thanks for adding the detailed tutorial, it overall looks good to me.
Do you plan to add the INT8/INT4 quantization in this PR or a separate PR later?

Thanks.

acceleration/fast_inference_tutorial/fast_inference_tutorial.ipynb
yiheng-wang-nv: Hi @Nic-Ma , to avoid restarting the notebook multiple times, it's recommended to use cli to run py script for getting benchmark results.

yiheng-wang-nv:
> acceleration/fast_inference_tutorial/fast_inference_tutorial.ipynb

Hi @Nic-Ma , thanks for the suggestion. I think we can consider adding quantization in a separate PR. Before adding it, it may need some time to:
1) prove it's faster
2) prove there will not have too much accuracy loss.

Nic-Ma:
Plan sounds good to me.

Thanks.

acceleration/fast_inference_tutorial/fast_inference_tutorial.ipynb
```diff
@@ -0,0 +1,646 @@
...
+    "\n",
+    "#### (Optional) Using the Python Script\n",
+    "\n",
+    "For convenience, a Python script, [`run_benchmark.py`](./run_benchmark.py), is available to run the benchmark. You can open a terminal and execute the following command to run the benchmark for all benchmark types:\n",
+    "\n",
+    "\n",
+    "```bash\n",
+    "for benchmark_type in \"original\" \"trt\" \"trt_gpu_transforms\" \"trt_gds_gpu_transforms\"; do\n",
+    "    python run_benchmark.py --benchmark_type \"$benchmark_type\"\n",
+    "done\n",
```
ericspod, line 402: You could instead put this into a cell with `%%bash` at the top to allow users to run command, or you could do it with Python more directly for those that don't have bash:

```python
for benchmark_type in ("original", "trt", "trt_gpu_transforms", "trt_gds_gpu_transforms"):
    !python run_benchmark.py --benchmark_type {benchmark_type}
```

ericspod: You should also state here that the script contains the same code as what's in this notebook, running it will generate a csv with the results for each type, but if the user wants to run the benchmark here in this notebook then can run the following cell with the commented lines uncommented.

yiheng-wang-nv:
> I've looked at the tutorial and it all looks good to me, however I am wondering about what the results show. It seems to me that GDS has the most impact so the example is just IO bound, using TRT or not has little impact. This is good to demonstrate how to overcome such issues, but it seems to me that the model is so small that it's not relevant to the benchmarks you're showing. If you used a much larger model with many more parameters the actual inference time itself would be significant. Since the inference results aren't considered you could just use a randomly initialised model so you don't need to load pre-trained weights. Thoughts?

Thanks @ericspod for the suggestions, I will use a more suitable model to show these features, and then update the PR

ericspod:
Hi @yiheng-wang-nv we'd like to get this tutorial through, do we have any progress on using a different model to demonstrate speedup better? Thanks!

PR #1951, opened 2025-03-09
Fix(2d_segmentation): Restore the predicted value of seg image to the range of 0~225 before saving it

Reqwey:
### Description
To fix the problem that evaluation output images are all black because `post_trans` scale their pixel values to 0~1.

### Checks


ericspod:
Thanks for the contribution @Reqwey! You can actually set a scaling factor for `SaveImage` with the [scale parameter](https://docs.monai.io/en/stable/transforms.html#monai.transforms.SaveImage). Setting `scale` to 255 should do the same as what you have here, so I'd suggest modifying your PR to do that. This is a bug in the tutorial either way so thanks for spotting it. Please also look at the instructions for resolving the DCO with a remedial commit.

Reqwey:
> Thanks for the contribution @Reqwey! You can actually set a scaling factor for `SaveImage` with the [scale parameter](https://docs.monai.io/en/stable/transforms.html#monai.transforms.SaveImage). Setting `scale` to 255 should do the same as what you have here, so I'd suggest modifying your PR to do that. This is a bug in the tutorial either way so thanks for spotting it. Please also look at the instructions for resolving the DCO with a remedial commit.

Thanks for your guidance! Done:
- [x] SaveImage(scale=255) (PR updated)
- [x] DCO remedial commit

It's my pleasure to be able to make contributions to this extraordinary project!

KumoLiu (approved):
Thanks for the fix, LGTM.

PR #1990, opened 2025-05-15
fix: handle metadata loading and shape calculation in transforms

Kirscher:
Fixes # .
handle metadata loading and shape calculation in transforms

### Description
- Add image_only=False and ensure_channel_first=True to LoadImaged transform to properly load metadata
- Fix shape mismatch in PreprocessAnisotropic.calculate_new_shape by handling 4D input shapes
- Addresses KeyError: 'image_meta_dict' and ValueError: operands could not be broadcast together

Note: The metadata loading issue might be related to MONAI version differences. 
Current version: 1.4.0

### Checks

### DIFF
--- a/modules/dynunet_pipeline/transforms.py
+++ b/modules/dynunet_pipeline/transforms.py
@@ -40,7 +40,7 @@ def get_task_transforms(mode, task_id, pos_sample_num, neg_sample_num, num_sampl
         keys = ["image", "label"]
     else:
         keys = ["image"]
-
+    # 1. loading
     load_transforms = [
         LoadImaged(keys=keys, image_only=False),
         EnsureChannelFirstd(keys=keys),
--- a/modules/dynunet_pipeline/transforms.py
+++ b/modules/dynunet_pipeline/transforms.py
@@ -42,7 +42,7 @@ def get_task_transforms(mode, task_id, pos_sample_num, neg_sample_num, num_sampl
         keys = ["image"]
 
     load_transforms = [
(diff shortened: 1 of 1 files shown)


modules/dynunet_pipeline/transforms.py
```diff
@@ -42,7 +42,7 @@ def get_task_transforms(mode, task_id, pos_sample_num, neg_sample_num, num_sampl
         keys = ["image"]
 
     load_transforms = [
-        LoadImaged(keys=keys),
+        LoadImaged(keys=keys, image_only=False, ensure_channel_first=True),
```
ericspod, line 45: If you have `ensure_channel_first` here then you can remove `EnsureChannelFirstd` in the sequence. You would use `EnsureChannelFirstd` next if you didn't have channel dimension information or wanted to do something before adding the dimension.

Kirscher: You are right the added `ensure_channel_first` is not required if you have `EnsureChannelFirstd` following. However, the `image_only=False` key is still needed in your `LoadImaged` transform to load the metadata and fix the error

ericspod: Sure that's true, so you should remove `ensure_channel_first=True` to be a clearer guide to using transforms correctly in this context.

Kirscher: Done 👍🏻

Kirscher:
DCO addressed, thanks

PR #2061, opened 2026-04-09
add opencv-python to requirements.txt

guozijn:
## Summary by CodeRabbit

* **Chores**
  * Updated project install-time dependencies to include OpenCV (opencv-python), ensuring the runtime now bundles OpenCV’s Python bindings.
  * Dependency-only change; no public interfaces or exported entities were modified.


ericspod (approved):
Hi @guozijn thanks for the update!

PR #2073, opened 2026-09-12
Fix NLST keypoint warping coordinates

### LINKED ISSUE #2021: learn2reg_nlst_paired_lung_ct.ipynb keypoints warping
Hi,
I'm working on a new registration method, and decided to use learn2reg_nlst_paired_lung_ct.ipynb as a base.
However, the conversion from keypoints from voxel coordinates to warping coordinates in the range [-1, 1] seems incorrect to me.
It is done like this (in the forward function):
```
offset = torch.as_tensor(fixed_image.shape[-3:]).to(fixed_keypoints.device) / 2
offset = offset[None][None]
ddf_keypoints = torch.flip((fixed_keypoints - offset) / offset, (-1,))
```

But in the Warp class of Monai, the conversion is done like this:
```
for i, dim in enumerate(grid.shape[1:-1]):
    grid[..., i] = grid[..., i] * 2 / (dim - 1) - 1
index_ordering: list[int] = list(range(spatial_dims - 1, -1, -1))
grid = grid[..., index_ordering]  # z, y, x -> x, y, z
```
Which seems correct to me, since voxel coordinates should max out at dim - 1  for indexing.

The two are almost equivalent, except that in the notebook, image.shape is used and not image.shape - 1, so I think the notebook is wrong. 
Feel free to correct me if I'm wrong, because I'm lacking info on what exactly are those keypoints and how they were obtained.

hmeixx:
Fixes #2021 

### Description
This PR fixes the keypoint coordinate normalization used when sampling the displacement field in `learn2reg_nlst_paired_lung_ct.ipynb`.

Previously, keypoint voxel coordinates were normalized using the image dimension:

```python
offset = torch.as_tensor(fixed_image.shape[-3:]).to(fixed_keypoints.device) / 2
```

This corresponds to `2 * x / dim - 1`, which does not map voxel indices `[0, dim - 1]` to `[-1, 1]`.

The normalization is updated to use `(dim - 1) / 2`, corresponding to:

```text
2 * x / (dim - 1) - 1
```

`align_corners=True` is also explicitly specified in `F.grid_sample` so that the sampling convention is consistent with this normalization and with MONAI `Warp`.

A small local `grid_sample` check confirmed that the updated normalization maps voxel centers correctly.

### Checks

Local PEP 8/format checks passed with runner.sh. Notebook execution was skipped by the existing runner skip pattern.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected keypoint displacement sampling to align with PyTorch’s corner-aligned coordinate system, improving registration accuracy and consistency.


ericspod (approved):

ericspod:
HI @hmeixx thanks for this fix, it looks good to me.

